PR #7, opened 2021-01-24
Encrypt collection name

Abhinav-grd:


lib/core/configuration.dart
```diff
@@ -108,7 +108,7 @@ class Configuration {
 
   String getHttpEndpoint() {
     if (kDebugMode) {
-      return "http://192.168.1.111";
+      return "http://2397bd9fe540.ngrok.io";
```
vishnukvmd, line 111: Revert pls?

lib/db/collections_db.dart
```diff
@@ -5,24 +5,35 @@ import 'package:path/path.dart';
...
 import 'package:photos/models/collection.dart';
 import 'package:sqflite/sqflite.dart';
+import 'package:sqflite_migration/sqflite_migration.dart';
 
 class CollectionsDB {
   static final _databaseName = "ente.collections.db";
-  static final _databaseVersion = 1;
-
   static final collectionsTable = 'collections';
+  static final collectionsTableCopy = 'collections_copy';
```
vishnukvmd, line 13: Why do we need this?

lib/db/collections_db.dart
```diff
@@ -36,28 +47,56 @@ class CollectionsDB {
...
-                  $columnUpdationTime TEXT NOT NULL
-                )
-                ''');
+  static List<String> addEncryptedName() {
+    return [
+      '''
+				ALTER TABLE $collectionsTable
+				ADD COLUMN $columnEncryptedName TEXT;
+      ''',
+      '''ALTER TABLE $collectionsTable
```
vishnukvmd, line 96: Maintain consistency in formatting?

lib/db/files_db.dart
```diff
@@ -7,14 +7,15 @@ import 'package:photos/models/file.dart';
 import 'package:path/path.dart';
 import 'package:sqflite/sqflite.dart';
 import 'package:path_provider/path_provider.dart';
+import 'package:sqflite_migration/sqflite_migration.dart';
 
 class FilesDB {
   static final _databaseName = "ente.files.db";
-  static final _databaseVersion = 1;
```
vishnukvmd, line 13: How is versioning maintained now?

lib/db/files_db.dart
```diff
@@ -54,21 +60,21 @@ class FilesDB {
...
+      '''
+          CREATE TABLE $tablename (
             $columnGeneratedID INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,
             $columnLocalID TEXT,
             $columnUploadedFileID INTEGER,
             $columnOwnerID INTEGER,
             $columnCollectionID INTEGER,
             $columnTitle TEXT NOT NULL,
-            $columnDeviceFolder TEXT NOT NULL,
+            $columnDeviceFolder TEXT,
```
vishnukvmd, line 77: Do you want to keep the `NOT NULL` constraint so that the `alterDeviceFolderToAllowNULL` makes sense?

lib/models/collection.dart
```diff
@@ -159,19 +175,22 @@ enum CollectionType {
 class CollectionAttributes {
   final String encryptedPath;
   final String pathDecryptionNonce;
-
+  final int version;
   CollectionAttributes({
     this.encryptedPath,
     this.pathDecryptionNonce,
+    this.version,
```
vishnukvmd, line 182: Set default value of `version` to `0` here?

lib/services/collections_service.dart
```diff
@@ -315,11 +324,38 @@ class CollectionsService {
   }
 
   String decryptCollectionPath(Collection collection) {
+    final key = collection.attributes.version != null
```
vishnukvmd, line 327: Keep the default value of `version` to `0` and compare with `0` instead?

lib/services/collections_service.dart
```diff
@@ -66,7 +66,9 @@ class CollectionsService {
         await setCollectionSyncTime(collection.id, null);
         FileRepository.instance.reloadFiles();
       } else {
-        updatedCollections.add(collection);
+        _cacheCollectionAttributes(collection);
```
vishnukvmd, line 69: This is anyway being done in L#77 right?

lib/services/collections_service.dart
```diff
@@ -66,7 +66,9 @@ class CollectionsService {
...
+        final updatedCollection = getcollectionWithDecryptedName(collection);
```
vishnukvmd, line 70: Why are we decrypting the name here? This would mean that the DB entry would also have the decrypted name. Wouldn't it be better to let `_cacheCollectionAttributes` have this logic?

lib/services/collections_service.dart
```diff
@@ -303,6 +310,8 @@ class CollectionsService {
         .then((response) {
       final collection = Collection.fromMap(response.data["collection"]);
       _cacheCollectionAttributes(collection);
+      final updatedCollection = getcollectionWithDecryptedName(collection);
+      _cacheCollectionAttributes(updatedCollection);
```
vishnukvmd, line 314: Again, I feel that `_cacheCollectionAttributes` can have the logic to decrypt and cache the decrypted name if `collection.name == null`. If yes, these two lines are unnecessary since we're already calling the same function in L#312.

lib/services/collections_service.dart
```diff
@@ -315,11 +324,38 @@ class CollectionsService {
...
+        ? getCollectionKey(collection.id)
+        : _config.getKey();
     return utf8.decode(CryptoUtil.decryptSync(
         Sodium.base642bin(collection.attributes.encryptedPath),
-        _config.getKey(),
+        key,
         Sodium.base642bin(collection.attributes.pathDecryptionNonce)));
   }
+
+  Collection getcollectionWithDecryptedName(Collection collection) {
```
vishnukvmd, line 336: Instead of this, just have a 
```dart
String _getDecryptedCollectionName(Collection collection)
```
That way a new `Collection` object does not have to be instantiated.

lib/db/files_db.dart
```diff
@@ -433,7 +458,7 @@ class FilesDB {
           $columnCreationTime=?''',
       whereArgs: [
         title,
-        deviceFolder,
+        deviceFolder ?? "",
```
vishnukvmd, line 461: Since we have removed the constraint from this column, we should be able to insert `NULL` values?

vishnukvmd:
git has detected [53 changed files](https://github.com/ente-io/frame/pull/7/files). Most of them are due to permission changes in the `thirdparty/flutter_image_compress` directory. Could you revert those?

lib/db/collections_db.dart
Abhinav-grd: There is no way to edit a table in SQLite , i have to make a new table and copy the previous table content to it.

lib/db/collections_db.dart
Abhinav-grd: image

Locally the file looks like this 😕

lib/db/files_db.dart
Abhinav-grd: It is done by the sqflite_migration library based on the number of migration scripts present.

lib/services/collections_service.dart
Abhinav-grd: will have to remove the final attribute from the collection name then

lib/db/files_db.dart
```diff
@@ -427,16 +457,26 @@ class FilesDB {
     int creationTime,
   ) async {
     final db = await instance.database;
-    final rows = await db.query(
-      table,
-      where: '''$columnTitle=? AND $columnDeviceFolder=? AND 
+    final rows = await (deviceFolder != null
```
vishnukvmd, line 460: Nitpicking, this line could read better. What if we extract `db.query(...)` into a variable, and do a 
```dart
final rows = await query;
``` 
instead?

lib/services/collections_service.dart
```diff
@@ -308,18 +313,47 @@ class CollectionsService {
   }
 
   void _cacheCollectionAttributes(Collection collection) {
+    _collectionIDToCollections[collection.id] = collection;
```
vishnukvmd, line 316: Isn't this a duplicate of L#321?

lib/services/collections_service.dart
```diff
@@ -308,18 +313,47 @@ class CollectionsService {
...
+    final updatedCollection = getCollectionWithDecryptedName(collection);
```
vishnukvmd, line 317: `updatedCollection` sounds misleading here. Call it `collectionWithDecryptedName` instead? Or forget about extracting it into a variable and directly do a 
```dart
_collectionIDToCollections[collection.id] = getCollectionWithDecryptedName(collection);
```

lib/services/collections_service.dart
```diff
@@ -308,18 +313,47 @@ class CollectionsService {
...
+        ? getCollectionKey(collection.id)
+        : _config.getKey();
     return utf8.decode(CryptoUtil.decryptSync(
         Sodium.base642bin(collection.attributes.encryptedPath),
-        _config.getKey(),
+        key,
         Sodium.base642bin(collection.attributes.pathDecryptionNonce)));
   }
+
+  Collection getCollectionWithDecryptedName(Collection collection) {
```
vishnukvmd, line 334: Prefix with `_` pls.

lib/services/collections_service.dart
```diff
@@ -308,18 +313,47 @@ class CollectionsService {
...
+    var name;
+    if (collection.encryptedName != null && collection.encryptedName != "") {
```
vishnukvmd, line 336: Instead of `collection.encryptedName != ""`, you can do a `collection.encryptedName.isNotEmpty`.

vishnukvmd (approved):

lib/services/collections_service.dart
Abhinav-grd: oops soory

lib/services/collections_service.dart
Abhinav-grd: i need to have the collection in the cache before calling `getCollectionDecryptedName` as it uses `getCollectionKey` which takes `collectionID` as parameter and finds the collection in the cache

vishnukvmd: What if we restructure code to something like this:
```dart
  Uint8List getCollectionKey(int collectionID) {
    if (!_cachedKeys.containsKey(collectionID)) {
      final collection = _collectionIDToCollections[collectionID];
      _cachedKeys[collectionID] = _getDecryptedKey(collection);
    }
    return _cachedKeys[collectionID];
  }

  Uint8List _getDecryptedKey(Collection collection) {
    final encryptedKey = Sodium.base642bin(collection.encryptedKey);
    if (collection.owner.id == _config.getUserID()) {
      return CryptoUtil.decryptSync(encryptedKey, _config.getKey(),
          Sodium.base642bin(collection.keyDecryptionNonce));
    } else {
      return CryptoUtil.openSealSync(
          encryptedKey,
          Sodium.base642bin(_config.getKeyAttributes().publicKey),
          _config.getSecretKey());
    }
  }
```

Then you could call `_getDecryptedKey(...)`?

lib/db/files_db.dart
```diff
@@ -7,14 +7,15 @@ import 'package:photos/models/file.dart';
...
+import 'package:sqflite_migration/sqflite_migration.dart';
 
 class FilesDB {
   static final _databaseName = "ente.files.db";
-  static final _databaseVersion = 1;
 
   static final Logger _logger = Logger("FilesDB");
 
   static final table = 'files';
+  static final temptable = 'temp_files';
```
vishnukvmd, line 18: *`tempTable`

lib/services/collections_service.dart
Abhinav-grd: Ya that's better

PR #15, opened 2021-07-05
delete token from server on logout

ua741:


lib/core/configuration.dart
```diff
@@ -103,6 +104,7 @@ class Configuration {
     }
     await _preferences.clear();
     await _secureStorage.deleteAll();
+    await UserService.instance.logout();
```
vishnukvmd, line 107: Let's await on this and verify the result before clearing `_preferences` or `_secureStorage`? Just to be safe?

vishnukvmd (approved):

lib/services/user_service.dart
```diff
@@ -90,6 +90,31 @@ class UserService {
...
+                headers: {
+                  "X-Auth-Token": _config.getToken(),
+                },
+              ));
+      if (response != null && response.statusCode == 200) {
+        await Configuration.instance.logout();
+        await dialog.hide();
+        Navigator.of(context).popUntil((route) => route.isFirst);
+      } else {
+        throw new Exception("log out action failed");
```
vishnukvmd, line 109: Would need to hide `dialog` here too? Also, could just log the exception here and show the generic error dialog.

ua741: > Would need to hide `dialog` here too? Also, could just log the exception here and show the generic error dialog.

As discussed offline, it's handled in the catch block. Ideally, we should never reach here.

PR #19, opened 2021-08-05
Live photos

ua741:


lib/ui/fading_app_bar.dart
```diff
@@ -249,21 +249,28 @@ class FadingAppBarState extends State<FadingAppBar> {
...
         (await getFile(file)).path,
         title: file.title,
-      )).id;
-    } else {
-      file.localID = (await PhotoManager.editor.saveVideo(
-        (await getFile(file)),
+      ));
+      file.localID = savedAsset.id;
+    } else if (file.fileType == FileType.video) {
+      final savedAsset = (await PhotoManager.editor.saveImageWithPath(
```
vishnukvmd, line 260: `saveVideo`?

lib/ui/fading_app_bar.dart
```diff
@@ -249,21 +249,28 @@ class FadingAppBarState extends State<FadingAppBar> {
...
     }
     await FilesDB.instance.insert(file);
     await LocalSyncService.instance.trackDownloadedFile(file);
     Bus.instance.fire(LocalPhotosUpdatedEvent([file]));
     await dialog.hide();
-    showToast("file saved to gallery");
+    if (file.fileType != FileType.livePhoto) {
+      showToast("file saved to gallery");
+    } else {
+      showToast("only image saved, live photo support coming soon");
```
vishnukvmd, line 273: If it's not too much effort, I feel we can give the user an option to choose between the photo and the video. Not sure when we'll come back to add this.

lib/ui/zoomable_live_image.dart
```diff
@@ -0,0 +1,273 @@
...
+    }
+    getFile(widget.photo, liveVideo: true).then((file) {
+      if (file != null && file.existsSync()) {
+        _logger.fine("loading  from local");
+        _setVideoPlayerController(file: file);
+      } else if (widget.photo.uploadedFileID != null) {
+        _logger.fine("loading from remote");
+        getFileFromServer(widget.photo, liveVideo: true).then((file) {
+          if (file != null && file.existsSync()) {
+            showToast("Found in remote remote");
```
vishnukvmd, line 148: No need for toast?

lib/ui/zoomable_live_image.dart
```diff
@@ -0,0 +1,273 @@
...
+    this.shouldDisableScroll,
+    @required this.tagPrefix,
+    this.backgroundDecoration,
+  }) : super(key: key);
+
+  @override
+  _ZoomableLiveImageState createState() => _ZoomableLiveImageState();
+}
+
+class _ZoomableLiveImageState extends State<ZoomableLiveImage>
```
vishnukvmd, line 38: Instead of managing all this state, do you want to use `ZoomableImageView` and `VideoWidget` on top of each other?

lib/utils/file_util.dart
```diff
@@ -101,54 +115,79 @@ Future<io.File> getFileFromServer(ente.File file,
...
+            fileExtension: fileExtension,
+          );
+          videoFile.deleteSync();
+        }
+      }
+    }
+    fileDownloadsInProgress.remove(file.uploadedFileID);
+    return liveVideo ? videoFileCache : imageFileCache;
+  }).catchError((e) {
+    fileDownloadsInProgress.remove(file.uploadedFileID);
```
vishnukvmd, line 178: Should we `rethrow` here?

lib/models/file.dart
```diff
@@ -61,6 +52,29 @@ class File {
...
   }
 
+  static FileType _fileTypeFromAsset(AssetEntity asset) {
+    FileType type = FileType.image;
+    switch (asset.type) {
+      case AssetType.image:
+        type = FileType.image;
+        // PHAssetMediaSubtype.photoLive.rawValue is 8
+        // This hack should go away once photos_manager support livePhotos
+        if (asset.subTypes != null &&
```
vishnukvmd, line 62: `kDebugMode` also?

lib/ui/zoomable_live_image.dart
```diff
@@ -0,0 +1,139 @@
...
+        autoPlay: true,
+        autoInitialize: true,
+        looping: false,
+        allowFullScreen: false,
+        showControls: false);
+    return Chewie(controller: _chewieController);
+  }
+
+  void _loadLiveVideo() {
+    if (_videoPlayerController != null) {
```
vishnukvmd, line 110: Wouldn't this always be `false` since it's only called once from `initState`??

lib/utils/file_download_util.dart
```diff
@@ -0,0 +1,65 @@
...
+import 'package:photos/core/configuration.dart';
+import 'package:photos/core/network.dart';
+import 'package:photos/models/file.dart' as ente;
+import 'package:photos/services/collections_service.dart';
+
+import 'crypto_util.dart';
+
+final _logger = Logger("file_download_util");
+
+Future<io.File> downloadAndDecrypt(ente.File file,
```
vishnukvmd, line 15: I'm skipping reviewing this assuming it's a direct copy-pasta.

lib/ui/zoomable_live_image.dart
```diff
@@ -0,0 +1,139 @@
...
+      return;
+    }
+    getFile(widget.photo, liveVideo: true).then((file) {
+      if (file != null && file.existsSync()) {
+        _logger.fine("loading  from local");
+        _setVideoPlayerController(file: file);
+      } else if (widget.photo.uploadedFileID != null) {
+        _logger.fine("loading from remote");
+        getFileFromServer(widget.photo, liveVideo: true).then((file) {
```
vishnukvmd, line 119: Doesn't `getFile` internally call `getFileFromServer`?

vishnukvmd: Okay I guess if it thinks it's locally available and then fails, we'll have to reload from server. We should think about placing this logic within `getFile` itself, so that the view layer can be agnostic to this. `todo` for future. 🤷

lib/utils/file_download_util.dart
ua741: Ack

vishnukvmd (approved):

PR #31, opened 2021-09-05
Android: verify permission before running local sync

ua741:


lib/services/local_sync_service.dart
```diff
@@ -49,11 +50,20 @@ class LocalSyncService {
...
 
-  Future<void> sync() async {
+  Future<void> sync({bool isAppInBackground = false}) async {
     if (!_prefs.containsKey(kHasGrantedPermissionsKey)) {
       _logger.info("Skipping local sync since permission has not been granted");
       return;
     }
+
+    if (Platform.isAndroid && !isAppInBackground) {
+      var permissionState = await PhotoManager.requestPermissionExtend();
```
vishnukvmd, line 60: `final`

lib/services/local_sync_service.dart
```diff
@@ -49,11 +50,20 @@ class LocalSyncService {
...
+      if (permissionState != PermissionState.authorized) {
+        _logger.severe(
+            "sync requested with invalid permission", permissionState);
```
vishnukvmd, line 63: Shouldn't we call `permissionState.toString()`? Not sure if it will get serialized otherwise.

vishnukvmd (commented):
If `isAppInBackground` is made an optional parameter, we might miss setting it correctly.

lib/services/local_sync_service.dart
ua741: > Shouldn't we call `permissionState.toString()`? Not sure if it will get serialized otherwise.
Will change it. 
fyi: currently it's getting logged as
```
      I   [LocalSyncService] [SEVERE] [2021-09-05 15:25:57.404439] sync requested with invalid permission
                         I  ⤷ type: PermissionState
                         I  ⤷ error: PermissionState.denied
```

ua741:
> If `isAppInBackground` is made an optional parameter, we might miss setting it correctly.
Switched to existing `_isBackground` flag in local sync service.

vishnukvmd (approved):

PR #35, opened 2021-09-10
Remove add/remove file and share collection action for incoming collection 

ua741:
* Hide add selected files to another collection button.
* Add folks to album & remove/delete from album is also not visible now
* ShareDialog: catch & log unhandled err  during collection path decryption err

### DIFF
--- a/lib/ui/gallery_app_bar_widget.dart
+++ b/lib/ui/gallery_app_bar_widget.dart
@@ -20,10 +20,10 @@ import 'package:photos/utils/toast_util.dart';
 enum GalleryAppBarType {
   homepage,
   local_folder,
+  // indicator for gallery view of collections shared with the user
   shared_collection,
-  collection,
-  search_results,
-  incoming_collection,
+  owned_collection,
+  search_results
 }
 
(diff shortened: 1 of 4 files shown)


lib/ui/gallery_app_bar_widget.dart
```diff
@@ -23,6 +23,7 @@ enum GalleryAppBarType {
   shared_collection,
   collection,
   search_results,
+  incoming_collection,
```
vishnukvmd, line 26: Perhaps you could rename `collection` to `owned_collection` for clarity? Also, use `shared_collection` instead of `incoming_collection`? I don't think it's being used anywhere right now.

ua741: Yes, `shared_collection` is only referred at once place but we are not setting it.  Will remove this particular reference of `share_collection` in the code here and reserve 'shared_collection` usage for incoming collections only.

https://github.com/ente-io/frame/blob/a0cbf3141e2d58f46605b50280225e817a24b268/lib/ui/gallery_app_bar_widget.dart#L197

vishnukvmd (approved):

PR #41, opened 2021-09-13
Support for moving files between albums

ua741:
Test Plan:
- [x]  Verified that move option is not visible on local folder, gallery view and incoming collections. 
- [x] Verified that files are getting moved, both on local device and web(secondary device)
- [x] File was previously present in destination but user removed it from the album. (Verified on both mobile & web)
- [x] File was already present in the destination album. (It removes file from source but doesn't create another copy in destination folder)

### DIFF
--- a/lib/services/collections_service.dart
+++ b/lib/services/collections_service.dart
@@ -12,6 +12,7 @@ import 'package:photos/core/network.dart';
 import 'package:photos/db/collections_db.dart';
 import 'package:photos/db/files_db.dart';
 import 'package:photos/events/collection_updated_event.dart';
+import 'package:photos/events/files_updated_event.dart';
 import 'package:photos/events/local_photos_updated_event.dart';
 import 'package:photos/models/collection.dart';
 import 'package:photos/models/collection_file_item.dart';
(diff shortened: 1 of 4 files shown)


lib/services/collections_service.dart
```diff
@@ -339,6 +340,61 @@ class CollectionsService {
...
+      options:
+          Options(headers: {"X-Auth-Token": Configuration.instance.getToken()}),
+    )
+        .then((value) async {
+      // insert files to new collection
+      await _filesDB.insertMultiple(files);
+      Bus.instance.fire(CollectionUpdatedEvent(toCollectionID, files));
+      //  todo: remove files from existing collection locally.
+      //  Ideally, remoteSync should take care of it.
+      Bus.instance.fire(CollectionUpdatedEvent(fromCollectionID, files,
```
ua741, line 375: Here, I intentionally didn't remove from local. Remove sync should remove the files from the collection. 
For immediate feedback, firing `CollectionUpdatedEvent` removes the files from the toCollection page if the user presses back.

vishnukvmd: This LGTM.

vishnukvmd (approved):

PR #47, opened 2021-09-19
[dart-null-safety] upgrade packages

ua741:
## Description
Upgrading packages to their null safe version.
Pending Packages:
image

Only [expansion_card](https://pub.dev/packages/expansion_card), [flutter_user_agent](https://pub.dev/packages/flutter_user_agent) & [super_logging](https://pub.dev/packages/super_logging) don't have a null-safe version. `flutter_user_agent` is discontinued, so will replace it. Will also explore options for expansion_card and super_logging.

## Test Plan
* Tested basic functionality of the app. 
* Went through readme of packages to verify no breaking changes is impacting frame app.

### DIFF
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -7,7 +7,7 @@ packages:
       name: animate_do
       url: "https://pub.dartlang.org"
     source: hosted
-    version: "1.7.5"
+    version: "2.0.0"
   archive:
     dependency: "direct main"
     description:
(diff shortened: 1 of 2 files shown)


vishnukvmd (approved):
We've cloned `super_logging`, and are not depending on the package directly.

ua741:
> We've cloned `super_logging`, and are not depending on the package directly.

Yes, same is true for image_compress and in_app_purchase. Luckily, both of these packages have null safe version available

PR #54, opened 2021-09-22
Add support for renaming own collections

ua741:
## Description
Ability to change name of an album

## Test Plan
- [x] Feature working as expected. 
- [x] Changed the name of screenshot collection and verified that the new screenshots are being uploaded on the same collection.

### DIFF
--- a/lib/ui/change_collection_name_dialog.dart
+++ b/lib/ui/change_collection_name_dialog.dart
@@ -1,0 +1,98 @@
+import 'package:flutter/cupertino.dart';
+import 'package:flutter/material.dart';
+import 'package:photos/services/user_service.dart';
+import 'package:photos/utils/dialog_util.dart';
+import 'package:photos/utils/email_util.dart';
+import 'package:photos/utils/toast_util.dart';
+
+class ChangeCollectionNameDialog extends StatefulWidget {
+  final String name;
+
+  const ChangeCollectionNameDialog({Key key, this.name}) : super(key: key);
+
+  @override
+  _ChangeCollectionNameDialogState createState() =>
+      _ChangeCollectionNameDialogState();
+}
+
+class _ChangeCollectionNameDialogState
+    extends State<ChangeCollectionNameDialog> {
+  String _newCollectionName;
+
+  @override
+  void initState() {
+    super.initState();
+    _newCollectionName = widget.name;
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return AlertDialog(
+      title: Text("enter new album name"),
+      content: SingleChildScrollView(
+        child: Column(
+          mainAxisAlignment: MainAxisAlignment.start,
+          crossAxisAlignment: CrossAxisAlignment.start,
+          children: [
+            TextFormField(
+              decoration: InputDecoration(
+                hintText: 'album name',
+                hintStyle: TextStyle(
+                  color: Colors.white30,
+                ),
+                contentPadding: EdgeInsets.all(12),
+              ),
+              onChanged: (value) {
+                setState(() {
+                  _newCollectionName = value;
+                });
+              },
+              autocorrect: false,
+              keyboardType: TextInputType.text,
+              initialValue: _newCollectionName,
+              autofocus: true,
+            ),
+          ],
+        ),
+      ),
+      actions: [
+        TextButton(
+          child: Text(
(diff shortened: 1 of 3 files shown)


lib/ui/gallery_app_bar_widget.dart
```diff
@@ -107,6 +115,37 @@ class _GalleryAppBarWidgetState extends State<GalleryAppBarWidget> {
...
+      return;
+    }
+    var result = await showDialog(
+      context: context,
+      builder: (BuildContext context) {
+        return ChangeCollectionNameDialog(name: appBarTitle);
+      },
+      barrierColor: Colors.black.withOpacity(0.85),
+    );
+    if (result == null) {
```
vishnukvmd, line 129: Why would result be `null`?

lib/ui/gallery_app_bar_widget.dart
```diff
@@ -107,6 +115,37 @@ class _GalleryAppBarWidgetState extends State<GalleryAppBarWidget> {
     );
   }
 
+  Future<dynamic> _renameAlbum(BuildContext context) async {
+    if (widget.type != GalleryAppBarType.owned_collection) {
+      return;
+    }
+    var result = await showDialog(
```
vishnukvmd, line 122: `final` by default

lib/ui/gallery_app_bar_widget.dart
```diff
@@ -54,6 +57,7 @@ class _GalleryAppBarWidgetState extends State<GalleryAppBarWidget> {
   final _logger = Logger("GalleryAppBar");
   StreamSubscription _userAuthEventSubscription;
   Function() _selectedFilesListener;
+  String appBarTitle;
```
vishnukvmd, line 60: `_appBarTitle` since it's a private variable.

lib/ui/change_collection_name_dialog.dart
```diff
@@ -0,0 +1,98 @@
...
+  @override
+  void initState() {
+    super.initState();
+    _newCollectionName = widget.name;
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return AlertDialog(
+      title: Text("enter new album name"),
```
vishnukvmd, line 31: `enter a new name`

lib/ui/change_collection_name_dialog.dart
```diff
@@ -0,0 +1,98 @@
...
+            ),
+          ),
+          onPressed: () {
+            if (_newCollectionName.trim().isEmpty) {
+              showErrorDialog(
+                  context, "empty name", "album name cannot be empty");
+              return;
+            }
+            if (_newCollectionName.trim().length > 100) {
+              showErrorDialog(context, "too large",
```
vishnukvmd, line 84: `name too large`

lib/ui/gallery_app_bar_widget.dart
```diff
@@ -107,6 +115,37 @@ class _GalleryAppBarWidgetState extends State<GalleryAppBarWidget> {
...
+      builder: (BuildContext context) {
+        return ChangeCollectionNameDialog(name: appBarTitle);
+      },
+      barrierColor: Colors.black.withOpacity(0.85),
+    );
+    if (result == null) {
+      return;
+    }
+
+    final dialog = createProgressDialog(context, "changing name...");
```
vishnukvmd, line 133: Could perhaps move this logic into `ChangeCollectionNameDialog` to reduce the logic on the app bar? Not sure.

lib/ui/gallery_app_bar_widget.dart
ua741: When user cancels the dialog or don't change the name, we are returning null to indicate no action is required.

lib/ui/gallery_app_bar_widget.dart
ua741: In future, during UX revamp, I was thinking of making this `ChangeCollectionNameDialog` a generic user input dialog, which can use reused. That's why I kept it separate for now.

lib/ui/gallery_app_bar_widget.dart
vishnukvmd: I thought I had deleted this comment.

vishnukvmd (approved):

PR #59, opened 2021-09-24
DB: delete duplicate files in collection & unique index to avoid this in future

ua741:
Title ^

## Test Plan
- Installed old version of the app and verified that duplicates entries where showing in a single collection
- Migrated to this version and verified all duplicate entries were gone.
- Verified that the magic metadata version was synced correctly.

### DIFF
--- a/lib/db/files_db.dart
+++ b/lib/db/files_db.dart
@@ -258,17 +258,20 @@ class FilesDB {
       '''
       DELETE from $table where ($columnCollectionID, $columnUploadedFileID) IN 
       (SELECT $columnCollectionID, $columnUploadedFileID from $table WHERE 
-        $columnCollectionID is not NULL and $columnUploadedFileID is NOT NULL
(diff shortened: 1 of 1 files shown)


lib/db/files_db.dart
```diff
@@ -252,6 +253,26 @@ class FilesDB {
     ];
   }
 
+  static List<String> addUniqueConstraintOnCollectionFiles() {
+    return [
+      '''
+      DELETE from $table where ($columnCollectionID, $columnUploadedFileID) IN 
+      (SELECT $columnCollectionID, $columnUploadedFileID from $table WHERE 
+        $columnCollectionID is not NULL and $columnUploadedFileID is NOT NULL
```
vishnukvmd, line 261: They've a default of `-1`, shouldn't we filter those out too?

vishnukvmd (approved):

PR #64, opened 2021-09-27
Fix handling of stripe subscription

ua741:
## Description
1. Stay on the stripe plan selection page during the onboarding step unless user  successfully subscribe via stripe or they click on `continue on free plan button` button.

2. Show payment details button even if current plan has expired

3. Bug fix: Show prompt to cancel non-stripe subscription only. This bug was stopping users from re-subscribing after the expiry of current stripe plan.

## Test Plan
Tested locally.

### DIFF
--- a/lib/ui/payment/stripe_subscription_page.dart
+++ b/lib/ui/payment/stripe_subscription_page.dart
@@ -85,17 +85,21 @@ class _StripeSubscriptionPageState extends State<StripeSubscriptionPage> {
   }
 
   FutureOr onWebPaymentGoBack(dynamic value) async {
-    if (widget.isOnboarding) {
+    // refresh subscription
+    await _dialog.show();
+    try {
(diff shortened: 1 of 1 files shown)


lib/ui/payment/stripe_subscription_page.dart
```diff
@@ -300,7 +304,8 @@ class _StripeSubscriptionPageState extends State<StripeSubscriptionPage> {
               if (isActive) {
                 return;
               }
-              if (!_isActiveStripeSubscriber &&
+              if (_currentSubscription.paymentProvider != kStripe &&
```
ua741, line 307: Previously: When last strip subscription expires, the app was not allowing the user to change plans/re-subscribe and showing this dialog: 'please cancel your existing subscription from stripe first`

vishnukvmd (approved):

PR #70, opened 2021-09-29
Fix bug which was resulting in repeated upload of same local file

ua741:
## Description
Issue:  Duplicate copy of local file was getting uploaded again & again

Possible Root cause: 
The app's db contains more than one entries of file with same title & device folder but certain file entries with localID as null.  

Previously we were only picking the `localID` from first entry from existing file. We may end up picking a file with localID as null & insert/update the file in DB with `localID` as null. Due to this,  during the next assets scan, we will again queue the device file for reupload.

## Test Plan
- [x] Reproduced this issue locally & verified the fix.
- [x] Shared internal build with bug reporter and verified that the fix works.

### DIFF
--- a/lib/services/remote_sync_service.dart
+++ b/lib/services/remote_sync_service.dart
@@ -218,15 +218,32 @@ class RemoteSyncService {
       final existingFiles = file.deviceFolder == null
           ? null
           : await _db.getMatchingFiles(file.title, file.deviceFolder);
-      if (existingFiles == null) {
-        // File uploaded from a different device
+      if (existingFiles == null || existingFiles.isEmpty) {
+        // File uploaded from a different device.
+        // Other rare possibilities : The local file is present on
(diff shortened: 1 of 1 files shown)


lib/services/remote_sync_service.dart
```diff
@@ -218,13 +218,26 @@ class RemoteSyncService {
...
+        // Note: The file.generatedID might be already set inside
+        // [DiffFetcher.getEncryptedFilesDiff]
+
+        final fileWithLocalID = existingFiles
+            .firstWhere((e) => e.localID != null, orElse: () => null);
+        if (fileWithLocalID != null) {
+          // File should ideally have the same localID
+          file.localID = fileWithLocalID.localID;
+        }
         file.localID = existingFiles[0]
```
ua741, line 241: will move this 🥺

vishnukvmd (approved):
Nice stuff.

PR #84, opened 2021-10-03
Add a button to archive / unarchive within the detail page

vishnukvmd:
## Description
^ Title

## Test Plan
- [x] Tested archive and unarchive from both action bar and detail page and ensured that feature is working as expected


lib/ui/fading_bottom_bar.dart
```diff
@@ -46,42 +48,77 @@ class FadingBottomBarState extends State<FadingBottomBar> {
...
+                },
+              ),
+            ),
+          ),
+        );
+      }
+      bool isArchived =
+          widget.file.magicMetadata.visibility == kVisibilityArchive;
+      children.add(
+        Tooltip(
```
ua741, line 86: We should hide this option for files shared with the user?

ua741 (approved):

PR #86, opened 2021-10-04
Fix bug during first load of remote livePhoto

ua741:
## Bug Description
During the load of live photo, we fetch both the image and video in parallel.  
For remote live photos, we were using the same download key for keeping track of both. This results in unexpected behaviour, where instead of resolving the future with image, we end up returning the video and the UI fails to load that. 
The subsequent load of the same file works fine as we fetch the correct file from the cache.

Fix:
Use separate IDs for tracking future for image and video for a live photo. 
## Test Plan

- Tested locally and verified that first load of remote image is working as expected.


lib/utils/file_util.dart
```diff
@@ -109,31 +109,53 @@ Future<io.File> getFileFromServer(
...
+
+Future<io.File> _getLivePhotoFromServer(ente.File file,
+    {ProgressCallback progressCallback, bool needLiveVideo}) async {
+  final downloadID = file.uploadedFileID;
+  try {
+    if (!livePhotoDownloadsTracker.containsKey(downloadID)) {
+      livePhotoDownloadsTracker[downloadID] =
+          _downloadLivePhoto(file, progressCallback: progressCallback);
+    }
+    final _livePhoto = await livePhotoDownloadsTracker[file.uploadedFileID];
```
vishnukvmd, line 145: `_` prefix is only used for instance variables, not local ones.

vishnukvmd (dismissed):

vishnukvmd:
When I open a live photo that hasn't been downloaded yet, and long press, it does not play even after the download is complete. If I switch to a different file and switch back and then long press, it plays as expected.

Ideal interaction would be to show a toast saying `downloading...` and then playback on long press once the download is complete.

lib/ui/zoomable_live_image.dart
```diff
@@ -95,37 +102,55 @@ class _ZoomableLiveImageState extends State<ZoomableLiveImage>
...
+    var videoFile = await getFile(widget.file, liveVideo: true)
+        .timeout(Duration(seconds: 15))
+        .onError((e, s) {
+      _logger.info("getFile failed ${_file.tag()}", e);
+      return null;
     });
+
+    if ((videoFile == null || !videoFile.existsSync()) &&
+        _file.isRemoteFile()) {
+      videoFile = await getFileFromServer(widget.file, liveVideo: true)
```
vishnukvmd, line 130: `getFile` internally calls `getFileFromServer` too right? Why are we retrying here?

ua741: This is to handle the case where we have a localID in the db but we fail to load the video from the local storage. (Saw similar pattern in ZoomableImageView widget).

vishnukvmd: Got it. It's to handle files that were deleted on local.

PR #96, opened 2021-10-08
Show network requests in debug build using alice

ua741:
## Description
By default, Alice will only cache last 1000 requests in memory.

## Test Plan
Tested on emulator


pubspec.yaml
```diff
@@ -105,6 +106,7 @@ dependencies:
 
 dependency_overrides:
   provider: ^6.0.1 # for chewie 1.2.2 https://github.com/brianegan/chewie/issues/530
+  flutter_local_notifications: ^8.1.1
```
vishnukvmd, line 109: Why do we need this?

lib/ui/settings/debug_section_widget.dart
```diff
@@ -21,6 +22,14 @@ class DebugSectionWidget extends StatelessWidget {
           child: SettingsTextItem(
               text: "key attributes", icon: Icons.navigate_next),
         ),
+        GestureDetector(
+          behavior: HitTestBehavior.translucent,
+          onTap: () async {
+            Network.instance.getAlice().showInspector();
```
vishnukvmd, line 28: Also show this on `longPress` for app-version?

lib/core/network.dart
```diff
@@ -1,27 +1,32 @@
...
 import 'package:package_info_plus/package_info_plus.dart';
 
 int kConnectTimeout = 15000;
 
 class Network {
   Dio _dio;
-
+  Alice _alice;
   Future<void> init() async {
+    _alice = Alice(darkTheme: true, showInspectorOnShake: true);
```
vishnukvmd, line 15: `showInspectorOnShake` might be scary in case of accidental shakes?

vishnukvmd (approved):

PR #140, opened 2021-11-12
Add support to show mnemonic recovery key

ua741:
## Description
Corresponding diff on bada-frame: https://github.com/ente-io/bada-frame/pull/232
## Test Plan

Verified that
 - Able to recover account on mobile using the mnemonic code (24 words)
-  Able to recover using the same 24 word code on web
- Same english word list is being used in the dart and js library

### DIFF
--- a/lib/ui/recovery_key_dialog.dart
+++ b/lib/ui/recovery_key_dialog.dart
@@ -1,10 +1,12 @@
 import 'dart:io' as io;
 import 'dart:ui';
 
+import 'package:bip39/bip39.dart' as bip39;
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 import 'package:flutter/widgets.dart';
 import 'package:photos/core/configuration.dart';
+import 'package:photos/core/constants.dart';
 import 'package:photos/utils/toast_util.dart';
 import 'package:share_plus/share_plus.dart';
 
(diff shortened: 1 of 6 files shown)


ua741:
This should not be landed before we add the support for same on web and test it cross device : generate on web -> validate on mobile and vice versa.

lib/ui/recovery_key_dialog.dart
```diff
@@ -136,6 +158,7 @@ class _RecoveryKeyDialogState extends State<RecoveryKeyDialog> {
     );
   }
 
+  // author civil bicycle drop gown ship file shallow chicken hub isolate among report piece cram crumble leopard guard spike segment update fall defy achieve
```
ua741, line 161: This key is for test account on local machine, so no harm done 😥 . Will remove it.

lib/ui/recovery_key_dialog.dart
```diff
@@ -128,6 +139,17 @@ class _RecoveryKeyDialogState extends State<RecoveryKeyDialog> {
...
+              CheckboxListTile(
+                  contentPadding: EdgeInsets.zero,
+                  controlAffinity: ListTileControlAffinity.trailing,
+                  title: Text('show human readable key'),
+                  value: _showMnenomicCode,
+                  onChanged: (value) {
+                    setState(() {
+                      _showMnenomicCode = value;
+                    });
+                  }),
```
ua741, line 152: Need to take a call whether we want to show the option to view non-mnemonic recovery key or not.
Accordingly we will either add similar option on web or remove it from the mobile.

vishnukvmd: Let's always show the mnemonic code, since the effort to write it down is lesser and that to save is the same.

vishnukvmd (approved):

vishnukvmd (approved):

PR #142, opened 2021-11-23
Sessions - view and terminate

vishnukvmd:
## Description
^ Title

## Test Plan
- [x] Manually verified that active sessions are listed correctly by the user agent
- [x] Manually verified that session termination works as expected

## Screenshots
photo_2021-11-24 01 53 31
photo_2021-11-24 01 53 29
photo_2021-11-24 01 53 26

### DIFF
--- a/lib/models/sessions.dart
+++ b/lib/models/sessions.dart
@@ -52,24 +52,33 @@ class Session {
   final String token;
   final int creationTime;
   final String ip;
-  final String userAgent;
+  final String ua;
+  final String prettyUA;
   final int lastUsedTime;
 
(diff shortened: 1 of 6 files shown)


ua741:
When I checked last time, we do have few sessions with null ip or user- agent. 
Also, for future & web, we can probably move this parsing for presentation on server itself..

vishnukvmd:
> When I checked last time, we do have few sessions with null ip or user- agent. Also, for future & web, we can probably move this parsing for presentation on server itself..

Done.

ua741 (approved):

PR #151, opened 2021-12-07
Remove files that have been deleted from the upload queue

vishnukvmd:
## Description
Earlier when files in the upload queue were deleted externally, the backup process would throw an error. This change handles this gracefully.

## Test Plan
None 😬

### DIFF
--- a/lib/utils/file_uploader.dart
+++ b/lib/utils/file_uploader.dart
@@ -16,6 +16,7 @@ import 'package:photos/core/event_bus.dart';
 import 'package:photos/core/network.dart';
 import 'package:photos/db/files_db.dart';
 import 'package:photos/db/upload_locks_db.dart';
+import 'package:photos/events/files_updated_event.dart';
 import 'package:photos/events/local_photos_updated_event.dart';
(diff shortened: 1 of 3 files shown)


lib/utils/file_uploader.dart
```diff
@@ -81,6 +82,19 @@ class FileUploader {
       }
       _pollBackgroundUploadStatus();
     }
+    Bus.instance.on<LocalPhotosUpdatedEvent>().listen((event) {
+      if (event.type == EventType.deletedFromDevice ||
+          event.type == EventType.deletedFromEverywhere) {
+        removeFromQueueWhere((file) {
+          for (final updatedFile in event.updatedFiles) {
+            if (file.generatedID == updatedFile.generatedID) {
```
abhinavkgrd, line 90: when is the `generatedID` added a file?

abhinavkgrd (approved):

lib/utils/file_uploader.dart
vishnukvmd: It's auto generated during insertion into the DB. So it unique per row.

PR #164, opened 2022-01-07
handle error while displaying sharingDialog

ua741:
## Description
Bug: In case of no internet, the progress dialog was never dismissed.

## Test Plan
Verified that the progress dialog was dismissed when there was no network.

### DIFF
--- a/lib/ui/gallery_app_bar_widget.dart
+++ b/lib/ui/gallery_app_bar_widget.dart
@@ -214,36 +214,34 @@ class _GalleryAppBarWidgetState extends State<GalleryAppBarWidget> {
     var collection = widget.collection;
     final dialog = createProgressDialog(context, "please wait...");
     await dialog.show();
-    if (collection == null) {
-      if (widget.type == GalleryAppBarType.local_folder) {
-        collection =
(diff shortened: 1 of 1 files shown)


lib/ui/gallery_app_bar_widget.dart
```diff
@@ -214,36 +214,34 @@ class _GalleryAppBarWidgetState extends State<GalleryAppBarWidget> {
...
-            _logger.severe(e, s);
-            await dialog.hide();
-            showGenericErrorDialog(context);
-          }
+    try {
+      if (collection == null) {
+        if (widget.type == GalleryAppBarType.local_folder) {
+          collection =
+              CollectionsService.instance.getCollectionForPath(widget.path);
+          collection ??=
```
vishnukvmd, line 222: Why the duplicate call?

ua741: I am not really sure, didn't touch the previous logic. After seeing the logic, I can see that we don't need line 220/21. Will remove it?

vishnukvmd (approved):

PR #174, opened 2022-01-23
Support for enabling/disabling public urls

ua741:

### DIFF
--- a/lib/services/collections_service.dart
+++ b/lib/services/collections_service.dart
@@ -286,6 +286,7 @@ class CollectionsService {
       collection.publicURLs?.add(PublicURL.fromMap(response.data["result"]));
       await _db.insert(List.from([collection]));
       _cacheCollectionAttributes(collection);
+      Bus.instance.fire(CollectionUpdatedEvent(collection.id, <File>[]));
     } on DioError catch (e) {
       if (e.response.statusCode == 402) {
         throw SharingNotPermittedForFreeAccountsError();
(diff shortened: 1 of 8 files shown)


vishnukvmd (approved):

lib/services/collections_service.dart
```diff
@@ -68,10 +72,20 @@ class CollectionsService {
     });
   }
 
+  int _getLastCollectionsSyncTime() {
+    if (!_prefs.containsKey(_collectionsSyncTimeKey)) {
+      _prefs.setBool(_collectionsPublicUrlReSync, true);
```
vishnukvmd, line 77: `await` since `setBool` is async.

lib/services/collections_service.dart
```diff
@@ -68,10 +72,20 @@ class CollectionsService {
...
+      return 0;
+    } else if (!_prefs.containsKey(_collectionsPublicUrlReSync)) {
+      _prefs.setBool(_collectionsPublicUrlReSync, true);
```
vishnukvmd, line 80: Here too.

lib/services/collections_service.dart
```diff
@@ -68,10 +72,20 @@ class CollectionsService {
     });
   }
 
+  int _getLastCollectionsSyncTime() {
```
vishnukvmd, line 75: Shouldn't we set the bools only after a successful collection sync?

vishnukvmd (approved):

PR #187, opened 2022-02-13
Remove blocking dialog for manually backed up files

ua741:

### DIFF
--- a/lib/db/files_db.dart
+++ b/lib/db/files_db.dart
@@ -578,6 +578,21 @@ class FilesDB {
     return _convertToFiles(results);
   }
 
+  // Files which user added to a collection manually but they are not uploaded yet.
+  Future<List<File>> getFilesUserUploadedManually() async {
+    final db = await instance.database;
+    final results = await db.query(
+      table,
+      where:
+          '($columnUploadedFileID IS NULL OR $columnUploadedFileID IS -1) AND '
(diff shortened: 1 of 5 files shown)


lib/db/files_db.dart
```diff
@@ -578,6 +578,21 @@ class FilesDB {
     return _convertToFiles(results);
   }
 
+  // Files which user added to a collection manually but they are not uploaded yet.
+  Future<List<File>> getFilesUserUploadedManually() async {
```
vishnukvmd, line 582: `getPendingUploads`?

lib/services/remote_sync_service.dart
```diff
@@ -212,6 +212,11 @@ class RemoteSyncService {
             " files were ignored for upload");
       }
     }
+    if (filesToBeUploaded.isEmpty) {
```
vishnukvmd, line 215: Should do this even if it's not empty?

lib/db/files_db.dart
ua741: `getPendingManualUploads`?

lib/services/remote_sync_service.dart
ua741: I was thinking of that, was just worried about performance hit. This felt okay to me, given the possible low usage of this flow.

vishnukvmd (approved):

vishnukvmd (approved):

PR #215, opened 2022-04-21
Redesign home states

ashilkn:

### DIFF
--- a/lib/ente_theme_data.dart
+++ b/lib/ente_theme_data.dart
@@ -83,3 +83,12 @@ ElevatedButtonThemeData buildElevatedButtonThemeData(
     ),
   ));
 }
+
+TextStyle gradientButtonTextTheme() {
+  return TextStyle(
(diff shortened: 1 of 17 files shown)


lib/ente_theme_data.dart
```diff
@@ -83,9 +80,18 @@ ElevatedButtonThemeData buildElevatedButtonThemeData(
...
+    padding: EdgeInsets.symmetric(vertical: 18),
     shape: const RoundedRectangleBorder(
       borderRadius: BorderRadius.all(Radius.circular(8)),
     ),
   ));
 }
+
+TextStyle gradientButtonTextTheme() {
+  return TextStyle(
+    color: Colors.white,
```
ua741, line 92: Will we need to change the color for the black and white theme? 
if yes, should we change this to Theme.of(context).textTheme.headline6.copyWith(..)?

ua741 (approved):

PR #222, opened 2022-04-28
redesigned 'albums' screen

ashilkn:


lib/utils/local_settings.dart
```diff
@@ -2,8 +2,9 @@ import 'package:shared_preferences/shared_preferences.dart';
 
 enum AlbumSortKey {
   albumName,
+  newestPhoto,
+  oldestPhoto,
   lastUpdated,
```
ua741, line 7: in enum, you would always want to add new options in the end. 
In old enum, index 1 is pointing to lastUpdated and in new one, it will point to newestPhoto.

ua741 (approved):

PR #230, opened 2022-05-05
Switch to trash diff v2

ua741:
## Description
Title ^

## Test Plan
Tested locally by deleting files from collection & trash.

### DIFF
--- a/lib/services/trash_sync_service.dart
+++ b/lib/services/trash_sync_service.dart
@@ -39,10 +39,9 @@ class TrashSyncService {
       _logger.fine("inserting ${diff.trashedFiles.length} items in trash");
       await _trashDB.insertMultiple(diff.trashedFiles);
     }
-    if (diff.deletedFiles.isNotEmpty) {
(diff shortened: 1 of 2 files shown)


lib/services/trash_sync_service.dart
```diff
@@ -68,12 +67,6 @@ class TrashSyncService {
         ignoredFiles.add(file);
       }
     }
-    for (TrashFile t in diff.deletedFiles) {
```
vishnukvmd, line 71: Why is this not needed any more?

lib/utils/trash_diff_fetcher.dart
```diff
@@ -27,13 +27,17 @@ class TrashDiffFetcher {
       );
       int latestUpdatedAtTime = 0;
       final trashedFiles = <TrashFile>[];
-      final deletedFiles = <TrashFile>[];
+      final deletedFilesUploadID = <int>[];
```
vishnukvmd, line 30: `deletedUploadIDs` will do?

vishnukvmd (approved):

PR #232, opened 2022-05-07
Redesign selection

ashilkn:
## Description

https://user-images.githubusercontent.com/77285023/167257225-758cf0be-3b06-4e2c-b438-ac7a28e656c0.mov


pubspec.lock
```diff
@@ -893,6 +893,15 @@ packages:
       url: "https://pub.dartlang.org"
     source: hosted
     version: "0.13.0"
+  pie_chart:
```
ua741, line 896: pie_chart is not needed, you can remove this section.

lib/ui/archive_page.dart
```diff
@@ -8,15 +8,18 @@ import 'package:photos/models/magic_metadata.dart';
 import 'package:photos/models/selected_files.dart';
 import 'package:photos/ui/gallery.dart';
 import 'package:photos/ui/gallery_app_bar_widget.dart';
+import 'package:photos/ui/gallery_overlay_widget.dart';
 
 class ArchivePage extends StatelessWidget {
   final String tagPrefix;
   final GalleryAppBarType appBarType;
+  final GalleryOverlayType overlayType;
```
ua741, line 16: Do we need another property called `GalleryOverlayType` or we can just use `GalleryAppBarType`. We can may be rename it to `GalleryType`?

ashilkn: Yes we can rename it to GalleryType

PR #242, opened 2022-05-13
Upgrade photos manger 

ua741:
## Test Plan
Verified that 
- [x] photos are getting imported correctly, including Live photos
- [x] Verified that edited files are getting uploaded, instead of original files
- [x] Request permissions (limited and full, and re-request on iOS in case of limited) is working as expected
- [x] Save video (download)
- [x] Save image (download and edit)
- [x] Delete


vishnukvmd:
Looks good. In addition to what you mentioned, these are other APIs we consume, that we should perhaps test for:
- Request permissions (limited and full, and re-request on iOS in case of limited)
- Change callbacks (to listen for changes)
- Clear cache (during app startup on iOS to clear application data that's used up while loading photos)
- Save video (download)
- Save image (download and edit)
- Delete

vishnukvmd (approved):

PR #243, opened 2022-05-13
Details screen remove dark/light theme

ashilkn:
The Details screen had changes according the the theme and it shouldn't have them. There were issues with the icons not being visible when the image is of full screen height and the AppBar had a solid color and some other issues.

https://user-images.githubusercontent.com/77285023/168264739-f572e687-e828-45a8-b59b-88598bc8002b.mov

This is the screen after redesign. Same for both dark and light theme.

after 2

I noticed that the text and icons of the image_editor_page on were not visible on light theme. Have fixed that.

after 1
(after redesign)

The image_editor_page can be optimised, will come back to it after a while.


lib/ui/trash_page.dart
```diff
@@ -1,30 +1,57 @@
...
+}
+
+class _TrashPageState extends State<TrashPage> {
+  Function() _selectedFilesListener;
+  @override
+  void initState() {
+    _selectedFilesListener = () {
+      setState(() {});
+    };
+    widget._selectedFiles.addListener(_selectedFilesListener);
```
ua741, line 41: Selecting any files in the trash page will refresh the entire page. Instead of that, can we create a separate widget which will be either empty or visible based on selected files?

PR #260, opened 2022-06-02
Progress bar

ashilkn:
## Description
The Progress bar wasn't working on the email entry screen and the OTT verification screen, so have made them working. It was not present in the Recovery key screen and the Subscription (on boarding) screen, have added it to them.
Screenshot from 2022-06-02 16-41-34
Screenshot from 2022-06-02 16-41-42

### DIFF
--- a/lib/ui/password_entry_page.dart
+++ b/lib/ui/password_entry_page.dart
@@ -478,6 +478,7 @@ class _PasswordEntryPageState extends State<PasswordEntryPage> {
             showAppBar: false,
             isDismissible: false,
             onDone: onDone,
+            showProgressBar: true,
           ));
     } catch (e) {
       _logger.severe(e);
--- a/lib/ui/payment/stripe_subscription_page.dart
+++ b/lib/ui/payment/stripe_subscription_page.dart
@@ -19,6 +19,7 @@ import 'package:photos/ui/progress_dialog.dart';
 import 'package:photos/ui/web_page.dart';
 import 'package:photos/utils/dialog_util.dart';
 import 'package:photos/utils/toast_util.dart';
+import 'package:step_progress_indicator/step_progress_indicator.dart';
 import 'package:url_launcher/url_launcher.dart';
 
 class StripeSubscriptionPage extends StatefulWidget {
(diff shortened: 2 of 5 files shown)


lib/ui/recovery_key_page.dart
```diff
@@ -51,22 +51,47 @@ class _RecoveryKeyPageState extends State<RecoveryKeyPage> {
     }
 
     return Scaffold(
-      appBar: widget.showAppBar
+      // appBar: widget.showAppBar
+      //     ? AppBar(
+      //         title: Text(widget.title ?? "Recovery key"),
+      //       )
+      //     : null,
```
ua741, line 58: remove this?

lib/ui/payment/stripe_subscription_page.dart
```diff
@@ -111,9 +112,26 @@ class _StripeSubscriptionPageState extends State<StripeSubscriptionPage> {
 
   @override
   Widget build(BuildContext context) {
-    final appBar = AppBar(
-      title: Text("Subscription"),
-    );
+    // final appBar = AppBar(
+    //   title: Text("Subscription"),
+    // );
+    final appBar = widget.isOnboarding
```
ua741, line 118: remove?

lib/ui/password_entry_page.dart
```diff
@@ -478,6 +478,7 @@ class _PasswordEntryPageState extends State<PasswordEntryPage> {
             showAppBar: false,
             isDismissible: false,
             onDone: onDone,
+            showProgressBar: true,
```
ua741, line 481: This page is also used for password reset/update. Do we want to set this true for all variants?

ashilkn:
Have removed the commented code

lib/ui/recovery_key_page.dart
```diff
@@ -29,7 +29,8 @@ class RecoveryKeyPage extends StatefulWidget {
       this.isDismissible,
       this.title,
       this.text,
-      this.subText})
+      this.subText,
+      this.showProgressBar = false})
```
ashilkn, line 33: It is set to false by default everywhere, so only when the optional parameter showProgressBar is set to true when RecoveryKeyPage() is called will the progress bar be shown.

PR #263, opened 2022-06-03
[Perf][AppStart] Don't await on featureFlag fetch from network

ua741:

### DIFF
--- a/lib/services/feature_flag_service.dart
+++ b/lib/services/feature_flag_service.dart
@@ -2,6 +2,7 @@ import 'dart:convert';
 import 'dart:io';
 
 import 'package:logging/logging.dart';
+import 'package:photos/core/constants.dart';
 import 'package:photos/core/network.dart';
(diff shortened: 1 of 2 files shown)


lib/services/feature_flag_service.dart
```diff
@@ -11,35 +13,52 @@ class FeatureFlagService {
   static final FeatureFlagService instance =
       FeatureFlagService._privateConstructor();
   static const kBooleanFeatureFlagsKey = "feature_flags_key";
+  FeatureFlags defaultFlags = FeatureFlags(
```
mnvr, line 16: Maybe this can be a static const on FeatureFlags class itself? e.g. `FeatureFlags.default` or `FeatureFlags.defaultFlags`.

mnvr (approved):
Otherwise looks this is a great improvement! Waiting on network requests during app init was risky... 😅

PR #264, opened 2022-06-03
Grand permission screen

ashilkn:
Before :
Screenshot from 2022-06-03 14-04-50

After:
Screenshot from 2022-06-03 15-39-34
Screenshot from 2022-06-03 15-38-20

### DIFF
--- a/lib/utils/toast_util.dart
+++ b/lib/utils/toast_util.dart
@@ -11,15 +11,16 @@ Future<void> showToast(String message,
     return Fluttertoast.showToast(
         msg: message,
         toastLength: toastLength,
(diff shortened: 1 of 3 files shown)


lib/utils/toast_util.dart
```diff
@@ -11,15 +11,15 @@ Future<void> showToast(String message,
     return Fluttertoast.showToast(
         msg: message,
         toastLength: toastLength,
-        gravity: ToastGravity.BOTTOM,
+        gravity: ToastGravity.TOP,
```
ua741, line 14: Do we want to move all toast to top?

ua741 (approved):

PR #274, opened 2022-06-06
Fix location attributes while converting Asset to File

ua741:
## Description
As per photos manager documentation:
```dart
/// Latitude value of the location when shooting.
  ///  * Android: `MediaStore.Images.ImageColumns.LONGITUDE`.
  ///  * iOS/macOS: `PHAsset.location.coordinate.longitude`.
  ///
  /// It's always null when the device is Android 10 or above.
  ///
  /// See also:
  ///  * https://developer.android.com/reference/android/provider/MediaStore.Images.ImageColumns#LATITUDE
  ///  * https://developer.apple.com/documentation/corelocation/cllocation?language=objc#declaration
  double? get longitude => _longitude;
```

```dart
/// Obtain latitude and longitude.
  ///  * Android: Obtain from `MediaStore` or EXIF (Android 10).
  ///  * iOS/macOS: Obtain from photos.
  ///
  /// [LatLng.latitude] and [LatLng.longitude] might be 0.
  Future<LatLng> latlngAsync() => plugin.getLatLngAsync(this);
```

### DIFF
--- a/lib/models/file.dart
+++ b/lib/models/file.dart
@@ -65,7 +65,8 @@ class File {
     file.localID = asset.id;
     file.title = asset.title;
     file.deviceFolder = pathName;
-    file.location = Location(asset.latitude, asset.longitude);
+    var latLng = await asset.latlngAsync();
+    file.location = Location(latLng.longitude, latLng.latitude);
     file.fileType = _fileTypeFromAsset(asset);
     file.creationTime = asset.createDateTime.microsecondsSinceEpoch;
     if (file.creationTime == 0) {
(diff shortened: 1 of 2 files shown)


lib/models/file.dart
```diff
@@ -65,7 +65,8 @@ class File {
     file.localID = asset.id;
     file.title = asset.title;
     file.deviceFolder = pathName;
-    file.location = Location(asset.latitude, asset.longitude);
+    var latLng = await asset.latlngAsync();
+    file.location = Location(latLng.longitude, latLng.latitude);
```
vishnukvmd, line 69: This impacts first time load performance significantly. Which is why we do this in `FileUploader` instead.

ua741: Oh. Got it. Will close this.

PR #300, opened 2022-06-11
Update launcher and splash screen icon

vishnukvmd:
- [x] Move from semi-bold to **bold**.
- [x] Reduce the size of the splash screen icon

photo_2022-06-11 18 52 58
photo_2022-06-11 18 52 54


mnvr (approved):
💪

PR #325, opened 2022-06-18
Add UI setting to keep device awake

ua741:
## Description

- https://roadmap.ente.io/prevent-sleep-p-4797/

We can add smartness regarding automatically enabling/disabling this lock in the future. That require more thoughts and testing to get it right.

### DIFF
--- a/lib/ui/settings/backup_section_widget.dart
+++ b/lib/ui/settings/backup_section_widget.dart
@@ -96,6 +96,26 @@ class BackupSectionWidgetState extends State<BackupSectionWidget> {
           ),
         ),
         SectionOptionDivider,
+        SizedBox(
+          height: 48,
+          child: Row(
+            mainAxisAlignment: MainAxisAlignment.spaceBetween,
+            children: [
+              Text(
+                "Keep device awake",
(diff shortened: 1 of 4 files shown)


vishnukvmd (dismissed):

lib/ui/settings/backup_section_widget.dart
```diff
@@ -103,7 +103,7 @@ class BackupSectionWidgetState extends State<BackupSectionWidget> {
             mainAxisAlignment: MainAxisAlignment.spaceBetween,
             children: [
               Text(
-                "Wakelock",
+                "Keep device awake",
```
mnvr, line 106: Alternative versions:

Disable screen lock

Disable automatic screen lock when ente is running?

This will ensure faster uploads by ensuring your device does not sleep when uploads are in progress.

Automatic screen lock disabled
Automatic screen lock enabled

mnvr: Versions 2:

Disable auto lock

Disable automatic screen lock when ente is running?

This will ensure faster uploads by ensuring your device does not sleep when uploads are in progress.

(no toast)

mnvr (approved):

PR #332, opened 2022-06-20
Fix swipe to go back interaction for iOS

ua741:
## Test Plan
- Verified that I am able to swipe right and go back previous screen.


mnvr (approved):
🔥

PR #339, opened 2022-06-22
Minor improvements to reduce battery usage

ua741:

### DIFF
--- a/lib/services/remote_sync_service.dart
+++ b/lib/services/remote_sync_service.dart
@@ -107,7 +107,6 @@ class RemoteSyncService {
         _existingSync = null;
       }
     } catch (e, s) {
-      _logger.severe("Error executing remote sync ", e, s);
       _existingSync.complete();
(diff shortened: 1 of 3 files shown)


lib/services/remote_sync_service.dart
```diff
@@ -268,6 +270,8 @@ class RemoteSyncService {
 
     if (toBeUploaded > 0) {
       Bus.instance.fire(SyncStatusUpdate(SyncStatus.preparing_for_upload));
+      // verify if the files can b uploaded before initiating actual upload.
+      await _uploader.canUpload(toBeUploaded);
```
vishnukvmd, line 274: `canUpload` doesn't do anything, so we could just expose `fetchURLs`, call that here with a comment on why we are executing it here?

vishnukvmd (approved):

vishnukvmd (approved):

PR #356, opened 2022-06-24
Show memory counter for large number of memories

ua741:


mnvr (approved):
@vishnukvmd we talked in Lounge, we'll just go with the indicator for now to at least fix the UI. Since there is a possibility that limiting the number of memories to the first 90 might break something elsewhere, and needs a bit more careful look to be sure.

PR #361, opened 2022-06-27
replaced the ente icon foreground with a high resolution one

ashilkn:


mnvr:
Screenshot 2022-06-27 at 1 28 03 PM

Let's remove this file?

ashilkn:
@mnvr  , sorry left if there accidentally. Will delete

PR #371, opened 2022-07-01
Organize UI Widgets

ua741:


lib/ui/common/common_elements.dart
```diff
@@ -0,0 +1,18 @@
+import 'package:flutter/material.dart';
+import 'package:flutter/widgets.dart';
+
+Widget nothingToSeeHere({Color textColor}) {
```
mnvr, line 4: I know this PR is just for reorganizing, so it's unrelated to this PR.

But we should consider creating Widgets instead of methods. It is a bit more typing, but it fits better with how Flutter wants to be structured, and is also more performant.

mnvr (approved):
LGTM otherwise!

mnvr (approved):

lib/ui/common/common_elements.dart
mnvr: (Doesn't have to be done in this PR, just resolve the conversation and merge please, apparently it won't let me merge until the conv is resolved)

ua741: This is already planned :).
Not just this, at places we are using final/static widgets, which were screaming (multiple rebuilds). Planning to soon fix these things iteratively.

mnvr: Awesome!

PR #379, opened 2022-07-05
Ensure sentry is not enabled for fdroid client

ua741:

### DIFF
--- a/lib/core/error-reporting/super_logging.dart
+++ b/lib/core/error-reporting/super_logging.dart
@@ -151,7 +151,8 @@ class SuperLogging {
     appVersion ??= await getAppVersion();
 
     final enable = config.enableInDebugMode || kReleaseMode;
-    sentryIsEnabled = enable && config.sentryDsn != null;
+    final isFDroidClient = await isFDroidBuild();
+    sentryIsEnabled = enable && config.sentryDsn != null && !isFDroidClient;
     fileIsEnabled = enable && config.logDirPath != null;
 
     if (fileIsEnabled) {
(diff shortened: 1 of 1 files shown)


mnvr (approved):

licaon-kter:
I've build a6681a2fc8181f2ce45fed6662e0518480187df7 and it still connects to `sentry-reporter.ente.io` fyi

licaon-kter:
Fyi, Netguard is on F-Droid ;)

ua741:
> Nope, I've build the F-Droid recipe at [a6681a2](https://github.com/ente-io/frame/commit/a6681a2fc8181f2ce45fed6662e0518480187df7)
> 
> Netguard has PER APP connection logs so you can selectively see which app does what ;)

Unfortunately, we were not able to reproduce it. If possible, please share a screenshot with some details about the request which was being made.

That said, we have pushed a [new version](https://github.com/ente-io/frame/tree/fdroid-0.6.12). This has more safegaurd checks to ensure that sentry is not configured for f-droid version. In this version, we have also removed the reference for `sentry-reporter.ente.io`.

licaon-kter:
No versionName update https://github.com/ente-io/frame/blob/fdroid-0.6.12/pubspec.yaml#L14

But anyway, 1da9410e8a5a7554eeca0877bf563d0e66ae4c8b is still connecting
index

licaon-kter:
_It's always DNS_

Looking again, it appears that Netguard shows the domain names that it knows, and since `api.`, `sentry.`, and `ente.io` share the same IPs it was showing me the first entry that it had cached.

So yes, this appears fixed.

There's also the fact Terms and Privacy policies connect to `sal.ente.io` which is actually `external.simpleanalytics.com` but that's not a direct issue.

PR #387, opened 2022-07-08
Enable self deletion for low-risk accounts

ua741:
## Test Plan
Tested locally

### DIFF
--- a/lib/models/delete_account.dart
+++ b/lib/models/delete_account.dart
@@ -1,0 +1,11 @@
+import 'package:flutter/foundation.dart';
+
+class DeleteChallengeResponse {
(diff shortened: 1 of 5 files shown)


mnvr:
The image to use - 
[doodles_rectangle_full.zip](https://github.com/ente-io/frame/files/9071262/doodles_rectangle_full.zip)

This image is a transparent PNG with a single color. Such images can be used as so called "template" images, and we can ask Flutter to fill them with some color at runtime. In this case, use Stroke/Light-2 or Stroke/Dark-2 depending on the mode – this way the same image will work in both modes.

mnvr (approved):

PR #389, opened 2022-07-09
Add view all device photos option for iOS

ua741:
## Description
In case of limited permissions, hopefully user will figure out that view all will only show photos for which they have granted permission :)


mnvr (dismissed):
Let's change the title to "On device". If you don't like that at all, we can also go with "All".

Otherwise lgtm!

mnvr (approved):

PR #395, opened 2022-07-12
Update illustration on the delete account page

vishnukvmd:


mnvr:
Before:



After:

mnvr (approved):
@vishnukvmd Is this okay? please merge if so (apparently Github doesn't allow to ask for a review from OP 😄)

PR #399, opened 2022-07-13
Redesign file info

ashilkn:
## Description
For video files, the duration was only shown for local files. Now it can be seen for every video file.
Before: 
[Screencast from 13-07-22 08:09:22 AM IST.webm](https://user-images.githubusercontent.com/77285023/178639333-82544978-3a61-41db-ac6a-7a30b74ba6d7.webm)

After : 
[Screencast from 13-07-22 07:56:14 AM IST.webm](https://user-images.githubusercontent.com/77285023/178638970-c0242856-67a6-41cf-83e0-a523a35f8db6.webm)

## Test Plan
Tested manually


lib/utils/date_time_util.dart
```diff
@@ -41,10 +42,12 @@ Map<int, String> _days = {
 };
 
 String getMonthAndYear(DateTime dateTime) {
+  //Jun 2022
```
ua741, line 45: Nice! Can you add these these above the method so it can act as quick documentation for the method? The IDE will suggest/hint the return format/

lib/utils/date_time_util.dart
```diff
@@ -53,13 +56,22 @@ String getDayAndMonth(DateTime dateTime) {
...
 String getDateAndMonthAndYear(DateTime dateTime) {
+  //30 Jun, 2022
   return dateTime.day.toString() +
       " " +
       _months[dateTime.month] +
       ", " +
       dateTime.year.toString();
 }
 
+// String getFullMonthAndDateAndYear(DateTime dateTime) {
```
ua741, line 67: remove this code?

lib/utils/date_time_util.dart
```diff
@@ -86,7 +99,13 @@ String getTime(DateTime dateTime) {
   return hours + ":" + minutes;
 }
 
+String getTimeIn12hrFormat(DateTime dateTime) {
+  //11:22 AM
+  return DateFormat.jm().format(dateTime);
+}
+
 String getFormattedTime(DateTime dateTime) {
+  //Thu, Jun 30, 2022 - 14:32
```
ua741, line 108: move above the method.. here and at other places too?

lib/ui/viewer/file/file_info_dialog.dart
```diff
@@ -437,4 +326,167 @@ class _FileInfoWidgetState extends State<FileInfoWidget> {
...
   }
+
+  Widget _getVideoDuration() {
+    if (widget.file.duration != 0) {
+      return Text(
+        secondsToHHMMSS(widget.file.duration),
+      );
+    }
+    return FutureBuilder(
+      future: widget.file.getAsset(),
```
ua741, line 337: file.getAsset() will not work when
- File is not present locally
- File is shared to ente app.

Also, we will need to hide the video information for live photos. 
We can may be add this check for now..
Call `_getVideoDuration` only when 
`(file.type == Video and (file.localID != null or file.duration !=0)`
And replace
`widget.file.getAsset()`
with `getFile()` (present in file_util)

ashilkn: For local files file.duration = 0. So file.getAsset() is only called for local files (see if condition on line 331). I think that is also the case for videos shared to ente app because it was working fine in the app when I ran it (the duration for shared videos). 

Yes, will hide video information for for live photos.

ua741: Sorry, we don't need to call getFile. Only need to call file.getAsset() for a local file, which not shared to the app.

ua741 (approved):

ua741 (approved):

PR #401, opened 2022-07-13
Android: Add support for System theme

ua741:
## Description
This is just a fallback for user who ask for this option till we finalize UI for tri-state option.

### DIFF
--- a/lib/app.dart
+++ b/lib/app.dart
@@ -2,6 +2,7 @@ import 'dart:io';
 
 import 'package:adaptive_theme/adaptive_theme.dart';
 import 'package:background_fetch/background_fetch.dart';
+import 'package:flutter/foundation.dart';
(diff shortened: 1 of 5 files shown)


mnvr:
Let me think, maybe we just fix this in one go.

ua741:
> Let me think, maybe we just fix this in one go.

Okay :). This is just so that we don't ask the user to uninstall and re-install. We can save brain cycle for now and think about this when we decide to redesign/improve settings screen.

mnvr:
Got it, you're right, I'll just try to do something quick

ashilkn (approved):

ashilkn:
The "Theme" heading on light theme has very low contrast with the background, other than that everything lgtm
temp

ashilkn (approved):

PR #413, opened 2022-07-25
Mobile search

ashilkn:
Search for collections
[Screencast from 25-07-22 07:36:09 PM IST.webm](https://user-images.githubusercontent.com/77285023/180797198-0d2fff58-321b-4445-a819-eb4eae33740e.webm)


lib/ui/status_bar_widget.dart
```diff
@@ -209,21 +211,35 @@ class StatusBarBrandingWidget extends StatelessWidget {
...
-      child: const Align(
-        alignment: Alignment.centerLeft,
-        child: Text(
-          "ente",
-          style: TextStyle(
-            fontWeight: FontWeight.bold,
-            fontFamily: 'Montserrat',
-            fontSize: 24,
-            height: 1,
+    FeatureFlagService.instance.enableSearchFeature();
```
ua741, line 214: Is this line needed?

lib/ui/viewer/search/collection_suggestions.dart
```diff
@@ -0,0 +1,117 @@
...
+
+class CollectionSuggestions {
+  final List<Collection> matchedCollections;
+  final BuildContext context;
+  const CollectionSuggestions(this.matchedCollections, this.context);
+
+  List<Widget> getSuggestions() {
+    List<Widget> collectionSuggestionWidgets = [];
+    collectionSuggestionWidgets =
+        generateSuggestionWidgets(collectionSuggestionWidgets);
```
ua741, line 20: Do we need to pass `collectionSuggestionWidgets` as parameter in `generateSuggestionWidgets`? If not, may be we can declare `    List<Widget> collectionSuggestionWidgets = [];` inside that method and return it here,

lib/services/collections_service.dart
```diff
@@ -863,6 +881,16 @@ class CollectionsService {
       }
     }
   }
+
+  Future<List<Collection>> getSearchedCollections(String query) async {
```
ua741, line 885: Instead of returning Collection, should we return `CollectionWithThumbnail` here? This way, you can simplify the rending of collections in the search result.

lib/ui/viewer/search/collection_suggestions.dart
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  List<Widget> generateSuggestionWidgets(
+    List<Widget> collectionSuggestionWidgets,
+  ) {
+    Future<List<File>> latestCollectionFiles =
+        CollectionsService.instance.getLatestCollectionFiles();
+    for (Collection collection in matchedCollections) {
+      CollectionWithThumbnail c;
+      collectionSuggestionWidgets.add(
```
ua741, line 31: Everything inside add, including `GestureDetector` should be part of a separate widget. `CollectionSearchResultWidget` or something similar. 
We can actually remove `CollectionSuggestions` completely.  Can discuss more about this on call/in-person if required.

lib/services/feature_flag_service.dart
```diff
@@ -83,6 +84,15 @@ class FeatureFlagService {
     }
   }
 
+  bool enableSearchFeature() {
+    String email = Configuration.instance.getEmail();
+    if (email.endsWith("@ente.io") || email == 'ashilkn99+ente77@gmail.com') {
```
ua741, line 89: email id can be null during sign up. Because we render the home widget with multiple pages, do verify that it's not breaking anything.

ua741, line 89: nit:
`return email != null && (endsWith("@ente.io") || kDebugMode)`

With `kDebugMode`, we can avoid hard coding personal email id.

lib/ui/status_bar_widget.dart
ashilkn: Nope, have removed it

lib/services/feature_flag_service.dart
ashilkn: Verified. Whenever enableSearchFeature() is called, email id is always != null.

lib/services/collections_service.dart
ashilkn: Done

lib/ui/viewer/search/collection_suggestions.dart
ashilkn: Done

ua741 (approved):

PR #415, opened 2022-07-28
[Part-0] Rewrite Sync: Use hash to avoid duplicate uploads

ua741:
#### Fixes redundant file update
Track local file with update modification time in a separate table and use file Hash to verify if the file content has actually changed or not.

## Test Plan

Tested locally for basic test cases.
- Favorite or Un-favorite a pic and verified the file update was not called.


lib/utils/file_uploader_util.dart
```diff
@@ -24,8 +26,20 @@ class MediaUploadData {
...
+  final String fileHash;
+  final String liveVideoHash;
+  final String zipHash;
 
-  MediaUploadData(this.sourceFile, this.thumbnail, this.isDeleted);
+  MediaUploadData(
+    this.sourceFile,
+    this.thumbnail,
+    this.isDeleted, {
+    this.fileHash,
```
vishnukvmd, line 39: See if we can have a single `hash` that can be deserialized?

lib/utils/file_uploader.dart
```diff
@@ -370,6 +318,19 @@ class FileUploader {
...
+        final isMappedToExistingUpload = await _mapToExistingUploadWithSameHash(
+          mediaUploadData,
+          file,
+          collectionID,
+        );
+        if (isMappedToExistingUpload) {
+          debugPrint(
+            "File success mapped to existing uploaded ${file.toString()}",
+          );
+          return file;
```
vishnukvmd, line 332: We should return the `uploadedFile`?

lib/utils/file_uploader.dart
```diff
@@ -476,23 +436,142 @@ class FileUploader {
...
+    duplicate files.
+  */
+  Future<bool> _mapToExistingUploadWithSameHash(
+    MediaUploadData mediaUploadData,
+    File fileToUpload,
+    int toCollectionID,
+  ) async {
+    if (fileToUpload.uploadedFileID != -1 &&
+        fileToUpload.uploadedFileID != null) {
+      _logger.warning('file is already uploaded, skipping mapping logic');
```
vishnukvmd, line 474: `error` so that we can catch it on Sentry?

lib/utils/file_uploader.dart
```diff
@@ -476,23 +436,142 @@ class FileUploader {
...
+      fileToUpload.fileType,
+      Configuration.instance.getUserID(),
+    );
+    if (existingFiles?.isEmpty ?? true) {
+      return false;
+    } else {
+      debugPrint("Found some matches");
+    }
+    // case a
+    File sameLocalSameCollection = existingFiles.firstWhere(
```
vishnukvmd, line 493: Variable could be `final`?

lib/utils/file_uploader.dart
```diff
@@ -476,23 +436,142 @@ class FileUploader {
...
+          element.localID == fileToUpload.localID,
+      orElse: () => null,
+    );
+    if (sameLocalSameCollection != null) {
+      debugPrint(
+        "sameLocalSameCollection: \n toUpload  ${fileToUpload.tag()} "
+        "\n existing: ${sameLocalSameCollection.tag()}",
+      );
+      // should delete the fileToUploadEntry
+      FilesDB.instance.deleteByGeneratedID(fileToUpload.generatedID);
```
vishnukvmd, line 506: Shouldn't we await?

lib/utils/file_uploader.dart
```diff
@@ -476,23 +436,142 @@ class FileUploader {
...
+      return true;
+    }
+
+    // case b
+    File fileMissingLocalButSameCollection = existingFiles.firstWhere(
```
vishnukvmd, line 511: `final`?

lib/utils/file_uploader.dart
```diff
@@ -476,23 +436,142 @@ class FileUploader {
...
+    );
+    if (existingFiles?.isEmpty ?? true) {
+      return false;
+    } else {
+      debugPrint("Found some matches");
+    }
+    // case a
+    File sameLocalSameCollection = existingFiles.firstWhere(
+      (element) =>
+          element.uploadedFileID != -1 &&
```
vishnukvmd, line 495: When we deserialize from the DB, I think we set it to `null`?

lib/utils/file_uploader.dart
```diff
@@ -476,23 +436,142 @@ class FileUploader {
...
+      );
+      // should delete the fileToUploadEntry
+      FilesDB.instance.deleteByGeneratedID(fileToUpload.generatedID);
+      return true;
+    }
+
+    // case b
+    File fileMissingLocalButSameCollection = existingFiles.firstWhere(
+      (element) =>
+          element.uploadedFileID != -1 &&
```
vishnukvmd, line 513: Same comment as (a)

lib/utils/file_uploader.dart
```diff
@@ -476,23 +436,142 @@ class FileUploader {
...
+      fileMissingLocalButSameCollection.localID = fileToUpload.localID;
+      await FilesDB.instance.insert(fileMissingLocalButSameCollection);
+      await FilesDB.instance.deleteByGeneratedID(fileToUpload.generatedID);
+      return true;
+    }
+
+    // case c and d
+    File fileExistsButDifferentCollection = existingFiles.firstWhere(
+      (element) =>
+          element.uploadedFileID != -1 &&
```
vishnukvmd, line 534: Same comment as (a) and (b)

lib/utils/file_uploader.dart
vishnukvmd: Also, if while fetching `existingFiles`, we are already checking for only uploaded files, we could ignore this check?

vishnukvmd (dismissed):

vishnukvmd (approved):

PR #424, opened 2022-08-13
Search improvements

vishnukvmd:


lib/ui/viewer/search/search_suggestions.dart
```diff
@@ -39,35 +39,37 @@ class SearchSuggestionsWidget extends StatelessWidget {
           borderRadius: const BorderRadius.all(Radius.circular(8)),
           child: Container(
             margin: const EdgeInsets.only(top: 6),
-            constraints: BoxConstraints(
-              maxHeight: MediaQuery.of(context).size.height * 0.5,
+            constraints: const BoxConstraints(
+              maxHeight: 324,
```
ashilkn, line 43: Wouldn't it be better to have a longer search results list in view for iPads/Tablets?

lib/ui/viewer/search/search_widget.dart
```diff
@@ -109,17 +111,20 @@ class _SearchWidgetState extends State<SearchWidget> {
...
               ),
             ),
-            results.isNotEmpty
-                ? SearchSuggestionsWidget(results)
-                : const NoResultWidget(),
+            _results.isNotEmpty
+                ? SearchSuggestionsWidget(_results)
+                : _query.isNotEmpty
+                    ? const NoResultWidget()
+                    : Container(),
```
ashilkn, line 127: const SizedBox.shrink() here instead of Container() will be better

lib/ui/viewer/search/search_suggestions.dart
vishnukvmd: As discussed, will keep this for now to deterministically cut the list view

ashilkn (approved):

PR #428, opened 2022-08-16
Holidays search

ashilkn:


lib/services/search_service.dart
```diff
@@ -132,26 +137,68 @@ class SearchService {
...
+  }
+
+  Future<List<HolidaySearchResult>> getHolidaySearchResults(
+    String query,
+  ) async {
+    final List<HolidaySearchResult> holidaySearchResult = [];
+
+    final nonCaseSensitiveRegexForQuery = RegExp(query, caseSensitive: false);
+
+    for (HolidayData holiday in allHolidays) {
```
vishnukvmd, line 176: Dart can infer the type when the right-hand side is defined. For example, here you could simply use `for (final holiday in allHolidays)`

lib/ui/viewer/search/search_widget.dart
```diff
@@ -136,31 +134,26 @@ class _SearchWidgetState extends State<SearchWidget> {
...
+    allResults.addAll(locationResults);
 
     return allResults;
   }
 
   bool isYearValid(int year) {
-    return year != null &&
-        year >= 1970 &&
-        year <= int.parse(DateTime.now().year.toString());
+    return year != null && year >= 1970 && year <= currentYear;
```
vishnukvmd, line 157: Unrelated to this PR, but this function can be private?

vishnukvmd (approved):

vishnukvmd (approved):

PR #431, opened 2022-08-17
Month search

ashilkn:


lib/services/search_service.dart
```diff
@@ -186,7 +188,29 @@ class SearchService {
...
-    return holidaySearchResult;
+    return holidaySearchResults;
+  }
+
+  Future<List<MonthSearchResult>> getMonthSearchResults(String query) async {
+    final List<MonthSearchResult> monthSearchResults = [];
+    final nonCaseSensitiveRegexForQuery = RegExp(query, caseSensitive: false);
+
+    for (var month in allMonths) {
+      if (month.name.startsWith(nonCaseSensitiveRegexForQuery)) {
```
vishnukvmd, line 199: If we are doing `startsWith`, do we need a regex?

lib/data/months.dart
```diff
@@ -0,0 +1,16 @@
+import 'package:photos/models/search/month_search_result.dart';
+
+List<MonthData> allMonths = [
+  MonthData('January', 1),
```
vishnukvmd, line 4: Should we stay consistent and start with 0 based indexing here?

vishnukvmd (approved):

PR #443, opened 2022-08-25
Fix GIF playback on iOS

vishnukvmd:
## Description
On iOS, `originFile` had to be loaded to render a GIF that would playback.

## Test Plan
- [x] Tested locally on iOS

### DIFF
--- a/lib/ui/viewer/file/zoomable_image.dart
+++ b/lib/ui/viewer/file/zoomable_image.dart
@@ -163,7 +163,7 @@ class _ZoomableImageState extends State<ZoomableImage>
       _loadingFinalImage = true;
       getFile(
         _photo,
-        isOrigin: isGIF(), // since playback only happens on origin files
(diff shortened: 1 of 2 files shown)


lib/ui/viewer/file/zoomable_image.dart
```diff
@@ -161,7 +163,11 @@ class _ZoomableImageState extends State<ZoomableImage>
 
     if (!_loadingFinalImage && !_loadedFinalImage) {
       _loadingFinalImage = true;
-      getFile(_photo).then((file) {
+      getFile(
```
ua741, line 166: Assuming that we don't need to delete the original file from the cache. App will eventually do it?

ua741 (dismissed):

lib/ui/viewer/file/zoomable_image.dart
vishnukvmd: Right, during next app startup

ua741 (approved):

PR #446, opened 2022-08-26
Hide albums

ashilkn:


lib/ui/viewer/gallery/gallery_app_bar_widget.dart
```diff
@@ -225,4 +256,7 @@ class _GalleryAppBarWidgetState extends State<GalleryAppBarWidget> {
       showGenericErrorDialog(context);
     }
   }
+
+  bool _isNotHiddenAndIsShared(Collection c) =>
+      (c.sharees.isNotEmpty || c.publicURLs.isNotEmpty) && !c.isArchived();
```
ua741, line 261: Similar to `isArchive`, we can define `isShared` method in the collection and use that. No need to this additional `_isNotHiddenAndIsShared` method.

lib/ui/viewer/gallery/gallery_app_bar_widget.dart
```diff
@@ -178,13 +177,45 @@ class _GalleryAppBarWidgetState extends State<GalleryAppBarWidget> {
...
             if (value == 2) {
-              await changeCollectionVisibility(
-                context,
-                widget.collection,
-                widget.collection.isArchived()
-                    ? kVisibilityVisible
-                    : kVisibilityArchive,
-              );
+              _isNotHiddenAndIsShared(widget.collection)
+                  ? showDialog(
```
ua741, line 181: Shall we use Choice dialog here?

var choice = await showChoiceDialog(
            context,
            "Hide shared album?",
            'Looks like you're trying to hide a shared album. The album will be hidden on your device, but it can still be seen by the people you have shared it with.',
            firstAction: 'Cancel',
            secondAction: 'Hide',
          );
          if(choice != DialogUserChoice.secondChoice) { 
          return;
           }

lib/models/collection.dart
```diff
@@ -47,6 +47,10 @@ class Collection {
     return mMdVersion > 0 && magicMetadata.visibility == kVisibilityArchive;
   }
 
+  bool isShared() {
+    return sharees.isNotEmpty || publicURLs.isNotEmpty;
```
ua741, line 51: can either of these fields be null?

lib/ui/viewer/gallery/sharing_hidden_album_warning_widget.dart
```diff
@@ -0,0 +1,52 @@
+import 'package:flutter/material.dart';
+import 'package:photos/ente_theme_data.dart';
+
+class SharingHiddenAlbumWarning extends StatelessWidget {
```
ua741, line 4: What was the need to creating separate alertDialog for this? Why can't we reuse the existing AlertDialog/ChoiceDialog etc?
If the design needs to be changed, may be we can create a generate Dialog with the new theme and just call that??

lib/models/collection.dart
ashilkn: Have added a null check. Not sure if they could be null, they can't be I think, but need more clarity on the codebase of this file.

lib/ui/viewer/gallery/sharing_hidden_album_warning_widget.dart
ashilkn: This is not a dialog, it is a banner on the hidden folder warning users they're sharing a hidden album. Have changed it to "SharingHiddenItemsBanner" as it now shows up if there are any shared files too.

ua741:
Closing this as this is not needed now. Trying to fix the current merge conflict and UI changes will take more time compared to starting fresh.

PR #454, opened 2022-08-30
Handle bugs related to no local authentication on device

ashilkn:


lib/ui/account/delete_account_page.dart
```diff
@@ -144,16 +145,19 @@ class DeleteAccountPage extends StatelessWidget {
...
-    AppLock.of(context).setEnabled(false);
-    const String reason = "Please authenticate to initiate account deletion";
-    final result = await requestAuthentication(reason);
-    AppLock.of(context).setEnabled(
-      Configuration.instance.shouldShowLockScreen(),
-    );
-    if (!result) {
-      showToast(context, reason);
-      return;
+    if (await LocalAuthentication().isDeviceSupported()) {
```
vishnukvmd, line 148: Let's extract this to a `LocalAuthenticationService`?

lib/utils/dialog_util.dart
```diff
@@ -26,15 +26,17 @@ ProgressDialog createProgressDialog(BuildContext context, String message) {
 
 Future<dynamic> showErrorDialog(
   BuildContext context,
-  String title,
-  String content,
-) {
+  String content, [
```
vishnukvmd, line 29: Changing the order of the arguments will break it for every caller

lib/services/user_service.dart
```diff
@@ -846,6 +849,26 @@ class UserService {
     }
   }
 
+  Future<bool> localAuthenticationService(
```
vishnukvmd, line 852: Better to pull this into a separate service, instead of adding it to UserService (since it doesn't have anything to do with the User)?

`LocalAuthenticationService.instance.requestAuthentication(context, reason)`?

ashilkn: Have named it requestLocalAuthentication as there is already another function called requestAuthentication in the same file.

lib/ui/settings/security_section_widget.dart
```diff
@@ -129,17 +125,25 @@ class _SecuritySectionWidgetState extends State<SecuritySectionWidget> {
...
               onChanged: (value) async {
-                AppLock.of(context).disable();
-                final result = await requestAuthentication(
-                  "Please authenticate to change lockscreen setting",
-                );
-                if (result) {
-                  AppLock.of(context).setEnabled(value);
-                  _config.setShouldShowLockScreen(value);
-                  setState(() {});
+                if (await LocalAuthentication().isDeviceSupported()) {
```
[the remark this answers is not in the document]
ashilkn: Could you explain why? The service would be something that most probably won't be used anywhere else. So I'm assuming you want to move it because there will be less lines of code on this block?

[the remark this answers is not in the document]
vishnukvmd: Yes, that, plus it'll be nice if all calls to the `LocalAuthentication` package are routed through a single proxy (`LocalAuthenticationService`). So if tomorrow we've to replace this dependency, changes would be contained to a single file.

vishnukvmd (approved):

PR #462, opened 2022-09-05
Fix missing feedback for remote sync

ua741:
## Description
Issues: 
- If a silent sync is in progress and a non-silent sync is request, then because of existingSync (already in progress), we do not update the status on UI.

- If a non-silent sync is requested, we only fire Status event once. Whenever this existing sync resumes (from bg to fg), then the UI might not render the correct status.

## Test Plan
YOLO'ing

### DIFF
--- a/lib/services/remote_sync_service.dart
+++ b/lib/services/remote_sync_service.dart
@@ -37,6 +37,7 @@ class RemoteSyncService {
   int _completedUploads = 0;
   SharedPreferences _prefs;
   Completer<void> _existingSync;
+  bool existingSyncSilent = false;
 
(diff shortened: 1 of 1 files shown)


lib/services/remote_sync_service.dart
```diff
@@ -37,6 +37,7 @@ class RemoteSyncService {
   int _completedUploads = 0;
   SharedPreferences _prefs;
   Completer<void> _existingSync;
+  bool existingSyncSilent = false;
```
vishnukvmd, line 40: `_` for private variables?

vishnukvmd (approved):

vishnukvmd (approved):

PR #480, opened 2022-09-13
Migrate to null safety

ashilkn:
Have manually tested eveything.


lib/models/ente_file.dart
```diff
@@ -1,14 +1,12 @@
...
-
 abstract class EnteFile {
   // returns cacheKey which should be used while caching entry related to
   // this file.
-  String cacheKey();
+  String? cacheKey();
 
   // returns localIdentifier for the file on the host OS.
   // Can be null if the file only exist on remote
-  String localIdentifier();
```
ashilkn, line 13: @ua741 this isn't being used anywhere as of now. It is being just overridden in file.dart and not used. Should it be removed?

lib/models/ente_file.dart
```diff
@@ -1,14 +1,12 @@
 // EnteFile is base file entry for various type of files
 // like DeviceFile,RemoteFile or TrashedFile
 
-// @dart=2.9
-
 abstract class EnteFile {
   // returns cacheKey which should be used while caching entry related to
   // this file.
-  String cacheKey();
+  String? cacheKey();
```
ua741, line 7: When will the `cacheKey` be null?

lib/models/ente_file.dart
ua741: yes, can remove it.

lib/models/file.dart
```diff
@@ -275,13 +275,9 @@ class File extends EnteFile {
   }
 
   @override
+  //add nullable return type when migrating to null safety
   String cacheKey() {
     // todo: Neeraj: 19thJuly'22: evaluate and add fileHash as the key?
     return localID ?? uploadedFileID?.toString() ?? generatedID?.toString();
```
ashilkn, line 281: @ua741 I made `cacheKey()` from `ente_file.dart` have a nullable return type because of this check. Could return null if every variable is null here. Also looks like this method is unused here, you have a `todo` comment on it, so should I keep it or remove it? cacheKey() can be non-nullable if this code is removed.

The return type must be the same type as (or a subtype of) the overridden method’s return type. So cacheKey() from `ente_file.dart` which is the overridden method must have a return type of `String?`

lib/models/ente_file.dart
ashilkn: @ua741 Have answered this on a comment on `file.dart`

ua741 (approved):

PR #491, opened 2022-09-16
Show Auto Logout Prompt when keys are missing but token is present

ua741:
## Test Plan
Tested locally on my device


mnvr (approved):

mnvr:
Note that this also updated the team ID (but that's fine, just noting that it is a change in this PR)

PR #495, opened 2022-09-16
Search: Support for search by calendar date & specific day

ua741:
## Test Plan
Tested locally.

### DIFF
--- a/lib/services/search_service.dart
+++ b/lib/services/search_service.dart
@@ -285,7 +285,6 @@ class SearchService {
       final int day = potentialDate.item1;
       final int month = potentialDate.item2.monthNumber;
       final int year = potentialDate.item3; // nullable
-
       final matchedFiles =
           await FilesDB.instance.getFilesCreatedWithinDurations(
         _getDurationsForCalendarDateInEveryYear(day, month, year: year),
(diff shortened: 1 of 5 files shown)


lib/services/search_service.dart
```diff
@@ -327,4 +368,52 @@ class SearchService {
...
+    final List<int> matchingYears = [];
+    if (parsedYear != null) {
+      bool foundMatch = false;
+      for (int i = searchStartYear; i <= currentYear; i++) {
+        if (i.toString().startsWith(parsedYear.toString())) {
+          matchingYears.add(i);
+          foundMatch = foundMatch || (i == parsedYear);
+        }
+      }
+      if (!foundMatch && parsedYear > 1000 && parsedYear <= currentYear) {
```
ashilkn, line 404: Could this be changed to >= searchStartYear ?

ua741: Can change..
Just thinking that for manual input, may be we should not add any limitation on the searchStartYear.  For partial matches, it make sense to keep the number of years and relevant query param/durations list to reasonable size.

People can have photos (scanned version of print) older than 1970 :)

PR #502, opened 2022-09-19
Migrate to null safety

ashilkn:
Locally tested, but not sure if sufficiently tested.


lib/models/collection.dart
```diff
@@ -1,28 +1,25 @@
...
 import 'package:photos/models/magic_metadata.dart';
 
 class Collection {
   final int id;
-  final User owner;
+  final User? owner;
   final String encryptedKey;
-  final String keyDecryptionNonce;
+  final String? keyDecryptionNonce;
   final String name;
```
ua741, line 11: name can be null afaik. At start, we used have name, later on we decided to store encrypted name

ashilkn: Have made it nullable

PR #508, opened 2022-09-21
Make colors const

ashilkn:


ua741 (approved):
PS: I have been making similar changes for colors while working on setting up base theme related stuff for figma components

PR #518, opened 2022-09-21
Migrate to null safety

ashilkn:
Tested locally

### DIFF
--- a/lib/models/encryption_result.dart
+++ b/lib/models/encryption_result.dart
@@ -1,15 +1,15 @@
 import 'dart:typed_data';
 
 class EncryptionResult {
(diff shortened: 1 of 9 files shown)


lib/models/encryption_result.dart
```diff
@@ -1,15 +1,15 @@
 import 'dart:typed_data';
 
 class EncryptionResult {
-  final Uint8List encryptedData;
-  final Uint8List key;
-  final Uint8List header;
-  final Uint8List nonce;
+  final Uint8List? encryptedData;
```
ua741, line 4: Ideally these fields in the result should never be empty

ua741 (approved):

PR #521, opened 2022-09-22
Migrate to null safety

ashilkn:
Tested locally.

### DIFF
--- a/lib/core/configuration.dart
+++ b/lib/core/configuration.dart
@@ -1,5 +1,3 @@
-// @dart=2.9
-
 import 'dart:convert';
 import 'dart:io' as io;
 import 'dart:typed_data';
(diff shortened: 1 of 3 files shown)


lib/models/file.dart
```diff
@@ -243,28 +243,6 @@ class File extends EnteFile {
...
-        o.generatedID == generatedID &&
-        o.uploadedFileID == uploadedFileID &&
-        o.localID == localID;
-  }
-
-  @override
-  int get hashCode {
-    return generatedID.hashCode ^ uploadedFileID.hashCode ^ localID.hashCode;
-  }
-
```
ua741, line 267: We need these as it's used when we perform contains check over a list of file

lib/core/configuration.dart
```diff
@@ -577,7 +578,7 @@ class Configuration {
 
   bool shouldHideFromRecents() {
     if (_preferences.containsKey(keyShouldHideFromRecents)) {
-      return _preferences.getBool(keyShouldHideFromRecents);
+      return _preferences.getBool(keyShouldHideFromRecents)!;
     } else {
       return false;
     }
```
ua741, line 584: nit: 
can change to `_preferences.getBool(keyShouldHideFromRecents) ?? false`

ua741 (approved):

PR #523, opened 2022-09-23
convert to getters

ashilkn:
Converted possible methods to getters and removed 'get' from the method names too as recommended here
https://dart.dev/guides/language/effective-dart/design#avoid-starting-a-method-name-with-get

Manual changes were only made to file.dart, all the other changes are due to changing method names in file.dart and those getting reflected in other files.

### DIFF
--- a/lib/models/file.dart
+++ b/lib/models/file.dart
@@ -116,7 +116,7 @@ class File extends EnteFile {
     return type;
   }
 
-  Future<AssetEntity?> getAsset() {
+  Future<AssetEntity?> get getAsset {
     if (localID == null) {
       return Future.value(null);
     }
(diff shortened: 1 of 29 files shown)


lib/models/file.dart
```diff
@@ -224,7 +224,7 @@ class File extends EnteFile {
     }
   }
 
-  String getDisplayName() {
+  String get displayName {
```
ua741, line 227: this can be null because title can be null

ashilkn: Have made sure that it doesn't return null
https://github.com/ente-io/frame/blob/f3213cc848ae6bdc447d0e4ce4e67f2740ffaf9f/lib/models/file.dart#L232

ua741 (approved):

PR #527, opened 2022-09-23
Migrate to null safety

ashilkn:


lib/services/feature_flag_service.dart
```diff
@@ -15,11 +13,11 @@ class FeatureFlagService {
 
   static final FeatureFlagService instance =
       FeatureFlagService._privateConstructor();
-  static const kBooleanFeatureFlagsKey = "feature_flags_key";
+  static const booleanFeatureFlagsKey = "feature_flags_key";
```
ua741, line 16: nit: can rename this to `_featureFagsKey`. AFAIK, this flag is not used outside this library

ua741 (approved):

PR #565, opened 2022-10-14
Do not add PopupMenuItem on home screen

ashilkn:


lib/ui/viewer/gallery/gallery_app_bar_widget.dart
```diff
@@ -253,33 +253,34 @@ class _GalleryAppBarWidgetState extends State<GalleryAppBarWidget> {
...
-            await _trashCollection();
-          } else if (value == 4) {
-            await _leaveAlbum(context);
-          } else {
-            showToast(context, "Something went wrong");
-          }
-        },
-      ),
-    );
+    if (widget.type != GalleryType.homepage) {
```
vishnukvmd, line 256: Should we call this function if it's `homepage`? Can we break out of this flow sooner?

PR #578, opened 2022-10-17
Fix apk update

ua741:
## Description
image

## Test Plan
- Locally changed the version to 6.39
- Verified that I am able to Set Us.
- Verified that Update Dialog successfully tried to install v0.6.40

### DIFF
--- a/lib/ui/settings/app_update_dialog.dart
+++ b/lib/ui/settings/app_update_dialog.dart
@@ -3,12 +3,12 @@
 import 'package:flutter/material.dart';
 import 'package:logging/logging.dart';
 import 'package:open_file/open_file.dart';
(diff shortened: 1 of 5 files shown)


lib/ui/settings/app_update_dialog.dart
```diff
@@ -34,20 +36,13 @@ class _AppUpdateDialogState extends State<AppUpdateDialog> {
       mainAxisSize: MainAxisSize.min,
       children: [
         Text(
-          widget.latestVersionInfo.name,
+          '${widget.latestVersionInfo.name} - what\'s new',
```
mnvr, line 39: Can we make it capital "W"

mnvr (approved):

PR #612, opened 2022-11-01
Storage card + backup screen fixes

ashilkn:


lib/ui/backup_settings_screen.dart
```diff
@@ -29,8 +29,13 @@ class BackupSettingsScreen extends StatelessWidget {
...
                 isSecondary: true,
-                onTap: () {
-                  Navigator.pop(context);
+                onTap: () async {
+                  int i = 0;
+                  //pop twice
+                  Navigator.popUntil(context, (route) {
+                    i++;
+                    return i == 3;
+                  });
```
ua741, line 38: Why not just call Pop twice instead of having this additional var :)

ua741 (approved):

ua741 (approved):

PR #627, opened 2022-11-08
Added logs to debug auto lock bug on video playback

ashilkn:


ua741:
As discussed, we should move the lock related logic inside VideoWidget. Also, on droid + iOS (where device lock is not enabled), we can let chewie handle this.

PR #640, opened 2022-11-13
Perf: Use cached value of collection key

ua741:
## Description

- Added extension method for chunking lists (needed for batching APIs)
- Helper method to emit time taken.

## Test Plan
After first run, collection refresh time went down from 60-80m to 15-20ms.

### DIFF
--- a/lib/services/collections_service.dart
+++ b/lib/services/collections_service.dart
@@ -21,6 +21,7 @@ import 'package:photos/events/collection_updated_event.dart';
 import 'package:photos/events/files_updated_event.dart';
 import 'package:photos/events/force_reload_home_gallery_event.dart';
 import 'package:photos/events/local_photos_updated_event.dart';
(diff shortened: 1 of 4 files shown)


lib/services/collections_service.dart
```diff
@@ -90,12 +91,14 @@ class CollectionsService {
   // within the collection.
   Future<List<Collection>> sync() async {
     _logger.info("Syncing collections");
+    final EnteWatch watch = EnteWatch("syncCollection")..start();
```
vishnukvmd, line 94: Should we remove the watch on production?

vishnukvmd (approved):

lib/services/collections_service.dart
ua741: Yes, I will will gradually remove them and by default turn off the logs.

PR #646, opened 2022-11-16
2fa toggle bug fix

ashilkn:


lib/core/configuration.dart
```diff
@@ -68,6 +69,8 @@ class Configuration {
   static const keyShouldShowLockScreen = "should_show_lock_screen";
   static const keyHasSelectedAnyBackupFolder =
       "has_selected_any_folder_for_backup";
+  static const keyHasEnabledTwoFactor = "has_enabled_two_factor";
```
ua741, line 72: Should we move this config key and logic insider UserService?  Configuration is already doing too many things which it should not be doing :)

lib/events/two_factor_status_change_event.dart
```diff
@@ -1,9 +1,10 @@
...
-
+import 'package:photos/core/configuration.dart';
 import 'package:photos/events/event.dart';
 
 class TwoFactorStatusChangeEvent extends Event {
   final bool status;
 
-  TwoFactorStatusChangeEvent(this.status);
+  TwoFactorStatusChangeEvent(this.status) {
+    Configuration.instance.setTwoFactor(value: status);
```
ua741, line 8: This doesn't seem like right place to do this :). Inside UserService,  when the call to enable or disable 2FA is completed, we can set this flag?

lib/services/user_service.dart
```diff
@@ -609,7 +609,7 @@ class UserService {
     }
   }
 
-  Future<void> setupTwoFactor(BuildContext context) async {
+  Future<void> setupTwoFactor(BuildContext context, Completer completer) async {
```
ua741, line 612: Passing completer here seems weird. Let's discuss if we really need this.

lib/events/two_factor_status_change_event.dart
ashilkn: Did this in a slightly different way. Attached a listener to to the `TwoFactorStatusChangeEvent` which will set the two factor status flag when the event is fired.

lib/services/user_service.dart
ashilkn: Will remove this in the future when the toggles are moved to a separate screen

lib/services/user_service.dart
```diff
@@ -49,6 +53,11 @@ class UserService {
   Future<void> init() async {
     emailValueNotifier =
         ValueNotifier<String>(Configuration.instance.getEmail());
+    _preferences = await SharedPreferences.getInstance();
+    setTwoFactor(fetchTwoFactorStatus: true);
```
ua741, line 57: Add .ignore?

ua741 (approved):

PR #649, opened 2022-11-18
Batch files

ashilkn:
## Description 

Batch selected files on move, add, remove, restore , archive, unarchive, hide and unhide.

## Test Plan

Tested locally

### DIFF
--- a/lib/services/collections_service.dart
+++ b/lib/services/collections_service.dart
@@ -933,21 +933,24 @@ class CollectionsService {
 
   Future<void> removeFromCollection(int collectionID, List<File> files) async {
     final params = <String, dynamic>{};
+    params["fileIDs"] = <int>[];
     params["collectionID"] = collectionID;
-    for (final file in files) {
-      if (params["fileIDs"] == null) {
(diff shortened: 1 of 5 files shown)


lib/services/collections_service.dart
```diff
@@ -719,35 +720,39 @@ class CollectionsService {
...
-      final key = decryptFileKey(file);
-      file.generatedID = null; // So that a new entry is created in the FilesDB
-      file.collectionID = collectionID;
-      final encryptedKeyData =
-          CryptoUtil.encryptSync(key, getCollectionKey(collectionID));
-      file.encryptedKey = Sodium.bin2base64(encryptedKeyData.encryptedData);
-      file.keyDecryptionNonce = Sodium.bin2base64(encryptedKeyData.nonce);
-      if (params["files"] == null) {
-        params["files"] = [];
+    params["files"] = [];
```
ua741, line 723: For readability, We can move inside the batch for loop? That way, we can avoid worrying about resetting/updating this in the finally block.

ua741 (approved):

PR #664, opened 2022-11-22
Minor improvements and fixes

ua741:
- Fix: Fetch 2FA status when account is configured
- Add delay in refreshing 2FA status on appStart


lib/ui/components/toggle_switch_widget.dart
```diff
@@ -60,7 +60,7 @@ class _ToggleSwitchWidgetState extends State<ToggleSwitchWidget> {
               activeColor: enteColorScheme.primary400,
               inactiveTrackColor: enteColorScheme.fillMuted,
               materialTapTargetSize: MaterialTapTargetSize.shrinkWrap,
-              value: toggleValue,
+              value: toggleValue ?? false,
```
ashilkn, line 63: `toggleValue` is non-nullable, why is this needed?

ua741: I was seeing null values in errors. Does it depend on some package/method which is not null safe?

ashilkn (approved):

PR #673, opened 2022-11-26
Removing Sharing for favorites + increase caption limit

ua741:
## Test Plan
Tested locally.

### DIFF
--- a/lib/ui/viewer/file/file_caption_widget.dart
+++ b/lib/ui/viewer/file/file_caption_widget.dart
@@ -14,6 +14,9 @@ class FileCaptionWidget extends StatefulWidget {
 
 class _FileCaptionWidgetState extends State<FileCaptionWidget> {
   static const int maxLength = 5000;
+  // counterThreshold represents the nun of char after which
+  // currentLength/maxLength will show up
+  static const int counterThreshold = 1000;
   int currentLength = 0;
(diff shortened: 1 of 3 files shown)


lib/ui/viewer/file/file_caption_widget.dart
```diff
@@ -13,7 +13,7 @@ class FileCaptionWidget extends StatefulWidget {
 }
 
 class _FileCaptionWidgetState extends State<FileCaptionWidget> {
-  int maxLength = 280;
```
mnvr, line 16: We also need to ellipsize on the photo screen, will that also happen with this change? Maybe that description widget on the photo screen have a max height already...

mnvr (approved):
LGTM otherwise, thank you

lib/ui/viewer/file/file_caption_widget.dart
```diff
@@ -13,7 +13,7 @@ class FileCaptionWidget extends StatefulWidget {
...
+  static const int maxLength = 5000;
```
mnvr, line 16: Also, let us start showing the counter text after 1000 characters (right now it seems to be after 100 characters, but with the new limit and 10 line height that's too less)

mnvr (approved):

PR #684, opened 2022-12-06
Advanced settings

vishnukvmd:
## Description
Added the ability to configure the gallery grid size

## Test Plan
- [x] Tested manually


lib/core/configuration.dart
```diff
@@ -560,6 +561,18 @@ class Configuration {
     }
   }
 
+  int getAlbumGridSize() {
+    if (_preferences.containsKey(keyAlbumGridSize)) {
```
ua741, line 565: Can move it to LocalPref, same place where we are storing sort order? Config is already too crowded

lib/ui/advanced_settings_screen.dart
```diff
@@ -0,0 +1,207 @@
...
+                  )
+                ],
+              ),
+            ),
+            Container(
+              height: 220.0,
+              color: const Color(0xfff7f7f7),
+              child: CupertinoPicker(
+                backgroundColor:
+                    Theme.of(context).backgroundColor.withOpacity(0.95),
```
ua741, line 179: at all places, can we please use enteText theme and ente Color scheme?

vishnukvmd: Done. Have used `backgroundElevated2` instead of `cupertinoPickerTopColor` because they look close enough (and I don't think this widget is on Figma yet)

ua741 (approved):

mnvr (approved):
Thank you 🙏

mnvr (approved):

PR #693, opened 2022-12-08
fix time upload

ua741:
- Android: Parse timestamp from fileName during upload
- Prefer existing creationTime if tileFromFileName is close to existing time

### DIFF
--- a/lib/models/file.dart
+++ b/lib/models/file.dart
@@ -188,12 +188,25 @@ class File extends EnteFile {
         creationTime = exifTime.microsecondsSinceEpoch;
       }
     }
-    // try to get the timestamp from fileName. In case of iOS, file names are
+    // Try to get the timestamp from fileName. In case of iOS, file names are
     // generic IMG_XXXX, so only parse it on Android devices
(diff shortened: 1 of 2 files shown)


lib/models/file.dart
```diff
@@ -188,12 +188,25 @@ class File extends EnteFile {
...
       final timeFromFileName = parseDateTimeFromFileNameV2(title!);
       if (timeFromFileName != null) {
-        creationTime = timeFromFileName.microsecondsSinceEpoch;
+        // only use timeFromFileName is the delta between current
+        // creationTime and timeFromFileName is larger than some threshold.
+        // This is done because many times the fileTimeStamp will only give us
+        // the date, not time value but the photo_manager's creation time will
+        // contain the time.
+        final bool useFileTimeStamp = creationTime == null ||
+            daysBetween(
```
vishnukvmd, line 202: If the diff is >= 2 we should `useFileTimeStamp`?

vishnukvmd: We could simplify the check to match the exact day?

PR #696, opened 2022-12-08
Action sheet + Large Button components

ashilkn:
### Description

Both these components are not fully complete yet, this is an intermediate PR.

All Buttons inside an ActionSheet : 

Simulator Screen Shot - iPhone 12 mini - 2022-12-08 at 18 34 37

Simulator Screen Shot - iPhone 12 mini - 2022-12-08 at 18 34 48

Simulator Screen Shot - iPhone 12 mini - 2022-12-08 at 19 02 39

Buttons under lightTheme outside of ActionSheet : 

Simulator Screen Shot - iPhone 12 mini - 2022-12-08 at 19 09 15

Some cases of ActionSheet :  

Simulator Screen Shot - iPhone 12 mini - 2022-12-08 at 19 14 12

Simulator Screen Shot - iPhone 12 mini - 2022-12-08 at 19 14 35

Simulator Screen Shot - iPhone 12 mini - 2022-12-08 at 19 15 43


lib/utils/add_separators_util.dart
```diff
@@ -0,0 +1,13 @@
+import 'package:flutter/material.dart';
+
+//This method returns a newly declared list with separators. It will not
+//modify the original list
+List<Widget> addSeparators(List<Widget> listOfWidgets, Widget separator) {
```
ua741, line 5: Let's move this logic inside the Build method? Generally we need to stop using functions which returns Widget.

When we need such abstraction, we can create widget which can handle such abstraction.

lib/ui/components/large_button_widget.dart
```diff
@@ -0,0 +1,313 @@
+import 'package:flutter/material.dart';
+import 'package:photos/theme/colors.dart';
+import 'package:photos/theme/ente_theme.dart';
+import 'package:photos/theme/text_style.dart';
+
+enum ButtonType {
```
ua741, line 6: define this enum outside of this widget class?
May be components/model?

lib/ui/components/large_button_widget.dart
```diff
@@ -0,0 +1,313 @@
...
+        buttonType == ButtonType.trailingIconPrimary ||
+        buttonType == ButtonType.trailingIconSecondary);
+  }
+
+  bool _isPrimary() {
+    return (buttonType == ButtonType.primary ||
+        buttonType == ButtonType.trailingIconPrimary);
+  }
+
+  bool _isSecondary() {
```
ua741, line 304: When methods could be extension Methods on the enum..
Inside the button_type.dart enum file, can add

```lang=dart
extension ExtentendButtonType on ButtonType {
    bool isPrimary() {
      return this == ButtonType.primary || ButtonType.trailingIconPrimary
   }
}
```

lib/ui/components/large_button_widget.dart
```diff
@@ -0,0 +1,313 @@
...
+    }
+    return null;
+  }
+
+  //Returning null to fallback to default color
+  TextStyle? _disabledLabelStyle(
+    EnteTextTheme textTheme,
+    EnteColorScheme colorScheme,
+  ) {
+    if (buttonType == ButtonType.primary ||
```
ua741, line 284: eval if this also can be extension method inside `extension ExtentendButtonType on ButtonType ` (more below)

lib/ui/components/large_button_widget.dart
ashilkn: Will do this. In newer dart versions this can be done using enhanced enums. Can have a getter like `isPrimary` inside the enum and need not create an extension
method for it.

lib/ui/components/large_button_widget.dart
ashilkn: Moved methods like these to inside the enum using enhanced enums.

PR #724, opened 2022-12-20
Tweak gradients

mnvr:
## Description

Tweak the gradients for the bottom left and right thumbnail icon overlays. Modify Figma to match these too.

These are not all the thumbnail overlay types, there are still some differences with Figma, e.g. the video thumbnail in Figma is different: 

Screenshot 2022-12-20 at 2 12 53 PM

But those are new types of overlays that can be picked up iteratively. This PR just modifies the existing ones, and also provides lower level abstractions like `_BottomRightOverlayIcon` to hopefully making adding the new ones easier.

## Test Plan

Tested on iOS simulator, with a variety of backgrounds and grid sizes.

Screenshot 2022-12-20 at 12 43 32 PM

Screenshot 2022-12-20 at 12 46 01 PM

Screenshot 2022-12-20 at 12 47 12 PM

Screenshot 2022-12-20 at 12 47 37 PM

Screenshot 2022-12-20 at 2 16 40 PM

### DIFF
--- a/lib/ui/viewer/file/file_icons_widget.dart
+++ b/lib/ui/viewer/file/file_icons_widget.dart
@@ -112,14 +112,30 @@ class FavoriteOverlayIcon extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    return const Align(
-      alignment: Alignment.bottomLeft,
-      child: Padding(
-        padding: EdgeInsets.only(left: 4, bottom: 4),
-        child: Icon(
-          Icons.favorite_rounded,
-          size: 20,
-          color: Colors.white, // fixed
+    return Container(
+      decoration: const BoxDecoration(
+        gradient: LinearGradient(
+          begin: Alignment.bottomLeft,
+          end: Alignment.center,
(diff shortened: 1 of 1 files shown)


lib/ui/viewer/file/file_icons_widget.dart
```diff
@@ -154,21 +127,142 @@ class TrashedFileOverlayText extends StatelessWidget {
...
-          color: fixedStrokeMutedWhite,
-        ),
-      ),
+    return LayoutBuilder(
+      builder: (context, constraints) {
+        double inset = 4;
+        double size = baseSize;
+
+        if (constraints.hasBoundedWidth) {
+          final w = constraints.maxWidth;
```
mnvr, line 162: This is one part which I'm not sure about. From documentation, and in testing, this seems to work, but I'm otherwise not certain if this value can be reliable always.

Although even if it is wrong, we'll not end up with too wrong thumbnails, just the sizes will be a bit off, so maybe the impact is bearable and we can fix it if we notice an issue.

ua741 (approved):

PR #725, opened 2022-12-20
Fix collection type

ua741:

### DIFF
--- a/lib/models/api/collection/create_request.dart
+++ b/lib/models/api/collection/create_request.dart
@@ -6,7 +6,7 @@ class CreateRequest {
   String keyDecryptionNonce;
   String encryptedName;
   String nameDecryptionNonce;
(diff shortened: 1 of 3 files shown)


lib/models/api/collection/create_request.dart
```diff
@@ -45,7 +45,7 @@ class CreateRequest {
     map['keyDecryptionNonce'] = keyDecryptionNonce;
     map['encryptedName'] = encryptedName;
     map['nameDecryptionNonce'] = nameDecryptionNonce;
-    map['type'] = type;
+    map['type'] = type.toString();
```
vishnukvmd, line 48: `toString` will still return the same since it's not overridden? Might have to call `typeToString(type)` instead

vishnukvmd (approved):

PR #751, opened 2022-12-30
UX: Enable pressedState for clearCache button

ua741:


mnvr (approved):

lib/ui/tools/debug/app_storage_viewer.dart
```diff
@@ -158,6 +158,8 @@ class _AppStorageViewerState extends State<AppStorageViewer> {
                             ),
                             menuItemColor:
                                 getEnteColorScheme(context).fillFaint,
+                            pressedColor:
+                                getEnteColorScheme(context).fillFaintPressed,
```
mnvr, line 162: Not a blocker for this PR, but would be nicer if MenuItem had this built in whenever it is acting as a button and can be interacted with. Just so that there is one place that can be changed, and we don't need to repeat this whenever we create such items.

---

### Details

Menu item can be used both as an action (e.g in this case the font is bold):

Screenshot 2022-12-30 at 11 38 05 AM

Or it can be an row of information (in which case the font is regular, and the menu item is not interactable)

Screenshot 2022-12-30 at 11 38 59 AM

There is also the case of a nested menu item where the font is not bold but the item is still interactable:

Screenshot 2022-12-30 at 11 40 24 AM

PR #753, opened 2022-12-30
NullSafety: Migrate PushService

ua741:

### DIFF
--- a/lib/services/push_service.dart
+++ b/lib/services/push_service.dart
@@ -1,5 +1,3 @@
-// @dart=2.9
-
 import 'package:firebase_core/firebase_core.dart';
 import 'package:firebase_messaging/firebase_messaging.dart';
 import 'package:logging/logging.dart';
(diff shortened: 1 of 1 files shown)


lib/services/push_service.dart
```diff
@@ -46,22 +44,24 @@ class PushService {
...
             kFCMTokenUpdationIntervalInMicroSeconds;
-    if (_prefs.getString(kFCMPushToken) != fcmToken ||
-        shouldForceRefreshServerToken) {
-      final apnsToken = await FirebaseMessaging.instance.getAPNSToken();
+    if ((_prefs.getString(kFCMPushToken) != fcmToken ||
+        shouldForceRefreshServerToken)) {
+      final String? apnsToken = await FirebaseMessaging.instance.getAPNSToken();
       try {
         _logger.info("Updating token on server");
         await _setPushTokenOnServer(fcmToken, apnsToken);
```
vishnukvmd, line 57: Shouldn't we do this only if both variables are non-null?

ua741: apns will be null for android

vishnukvmd (approved):

PR #763, opened 2023-01-03
Un-uploaded duplicate on device folder

ashilkn:
## Description

Bug : If a file in a backed-up device folder is deleted from the device folder (and the device) and a file which existed in the device folder is downloaded to the same device folder from the backed-up device folder (which has the files from before deletion), on the device folder, the downloaded file is marked as not backed-up while the file has always been present in the backed-up device folder. What is expected is for the file to come up on the device folder without the not backed-up icon over its thumbnail. This PR has resolved this bug.

## Test Plan

Tested locally.


ua741 (approved):

lib/services/local_sync_service.dart
```diff
@@ -258,12 +258,6 @@ class LocalSyncService {
     }
   }
 
-  Future<void> trackDownloadedFile(String localID) async {
-    final downloadedIDs = _getDownloadedFileIDs();
-    downloadedIDs.add(localID);
-    await _prefs.setStringList(kDownloadedFileIDsKey, downloadedIDs);
-  }
-
   List<String> _getDownloadedFileIDs() {
```
vishnukvmd, line 261: We could remove this as well? @ua741

ua741 (dismissed):

lib/services/local_sync_service.dart
ua741: Yes

vishnukvmd: @ashilkn

lib/ui/viewer/file/fading_app_bar.dart
```diff
@@ -411,6 +411,9 @@ class FadingAppBarState extends State<FadingAppBar> {
   Future<void> _download(File file) async {
     final dialog = createProgressDialog(context, "Downloading...");
     await dialog.show();
+    //Disabling notifications for assets changing to insert the file into
+    //files db before triggering a sync.
+    PhotoManager.stopChangeNotify();
```
ua741, line 416: Let's disable if after getFile() is done?

lib/ui/viewer/file/fading_app_bar.dart
```diff
@@ -468,6 +461,9 @@ class FadingAppBarState extends State<FadingAppBar> {
       _logger.warning("Failed to save file", e);
       await dialog.hide();
       showGenericErrorDialog(context: context);
+    } finally {
+      PhotoManager.startChangeNotify();
+      LocalSyncService.instance.checkAndSync();
```
ua741, line 466: I am wondering if we really need to sync files here.. May be Just running `_refreshDeviceFolderCountAndCover` might be good enough so that on device albums are in correct state.

Anyways, after thinking more about it, it's fine i guess. Please add .ignore() at the end.

lib/services/local_sync_service.dart
```diff
@@ -409,16 +384,20 @@ class LocalSyncService {
...
-      } else {
-        sync().then((value) => _refreshDeviceFolderCountAndCover());
-      }
+      checkAndSync();
     });
     PhotoManager.startChangeNotify();
   }
+
+  Future<void> checkAndSync() async {
+    _logger.info("Something changed on disk");
```
ua741, line 393: This log line should be inside `addChangeCallback`

lib/ui/viewer/file/fading_app_bar.dart
ua741: Also, let's only call this force sync when getFIle returns a file. If the File download fails, there's no need to trigger the sync.
Ideally, we should call it when both fileDownload and saveAsset is done without error, but adding a check just for file download might be easy & good enough.

ashilkn: We do need to sync for the device album to be in the correct state. Just running `_refreshDeviceFolderCountAndCover` is only updating the thumbnail of the device folder is what I observed while testing.

ua741 (approved):

PR #782, opened 2023-01-07
Switch to async variants

ua741:
- Use async variante instead of sync
- ThumbnailUtil: switch to async methods

### DIFF
--- a/lib/utils/crypto_util.dart
+++ b/lib/utils/crypto_util.dart
@@ -34,9 +34,9 @@ Uint8List cryptoPwHash(Map<String, dynamic> args) {
   );
 }
 
-Uint8List cryptoGenericHash(Map<String, dynamic> args) {
+Future<Uint8List> cryptoGenericHash(Map<String, dynamic> args) async {
   final sourceFile = io.File(args["sourceFilePath"]);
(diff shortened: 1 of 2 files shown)


vishnukvmd (approved):

lib/utils/crypto_util.dart
```diff
@@ -34,9 +34,9 @@ Uint8List cryptoPwHash(Map<String, dynamic> args) {
   );
 }
 
-Uint8List cryptoGenericHash(Map<String, dynamic> args) {
+Future<Uint8List> cryptoGenericHash(Map<String, dynamic> args) async {
```
vishnukvmd, line 37: The caller for this should now await?

ua741: This is the only place where it's being called. I am not sure if we need to change anything else. Fwiw, we are passing async method to computer inside local_sync_util.
```code=dart
static Future<Uint8List> getHash(io.File source) {
    return _computer.compute(
      cryptoGenericHash,
      param: {
        "sourceFilePath": source.path,
      },
    );
  }
```

vishnukvmd (approved):

PR #786, opened 2023-01-07
Replace choice dialogs

ashilkn:
## Description

- The ButtonWidget can now show a success state even if it's `onTap()` takes less than debounce time to execute if `shouldShowSuccessConformation` is passed as true.
- Replaced old dialogs with new ones.

https://user-images.githubusercontent.com/77285023/211150047-46bfabee-1b22-4553-a7eb-e19e3f41e651.mp4

https://user-images.githubusercontent.com/77285023/211150054-c1da5ff2-597c-41a5-9370-94a6132ca87c.mp4

## Test Plan

Tested Locally.

### DIFF
--- a/lib/ui/account/password_reentry_page.dart
+++ b/lib/ui/account/password_reentry_page.dart
@@ -81,7 +81,7 @@ class _PasswordReentryPageState extends State<PasswordReentryPage> {
             _logger.severe("Password verification failed", e, s);
             await dialog.hide();
             final dialogChoice = await showNewChoiceDialog(
-              context: context,
+              context,
               title: "Recreate password",
               body: "The current device is not powerful enough to verify your "
                   "password, so we need to regenerate it once in a way that "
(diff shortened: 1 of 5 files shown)


lib/ui/account/password_reentry_page.dart
```diff
@@ -80,19 +80,16 @@ class _PasswordReentryPageState extends State<PasswordReentryPage> {
           } on KeyDerivationError catch (e, s) {
             _logger.severe("Password verification failed", e, s);
             await dialog.hide();
-            final dialogUserChoice = await showChoiceDialog(
-              context,
-              "Recreate password",
-              "The current device is not powerful enough to verify your "
+            final dialogChoice = await showNewChoiceDialog(
```
ua741, line 83: In general, context can be always the first parameter, which is not optional. Adding it as required name parameter increases the verbosity and it might be really need to be named parameter because there will be most of the time just one input param of type context.

ua741 (approved):

ua741 (approved):

PR #815, opened 2023-01-17
Set the zoom-scale correctly once full-res image has loaded

vishnukvmd:


lib/ui/viewer/file/zoomable_image.dart
```diff
@@ -101,11 +113,17 @@ class _ZoomableImageState extends State<ZoomableImage>
   }
 
   void _loadNetworkImage() {
+    if (_loadingFinalImage) {
+      return;
+    } else {
+      _loadingFinalImage = true;
+    }
```
ua741, line 120: The state management is little complex to understand in this widget.
Do you want to double check `_loadNetworkImage` method for remote only widget and test/reason out yourself that this change won't cause any regression? This doesn't look 100% safe change to me.

ua741: May be check check should before calling `getFileFromServer` and set the flag accordingly (similar to loading image from local?)

vishnukvmd: So this was a bug that's unrelated to the PR. What's happening is that `_loadNetworkImage` is called from `build`, and each time we set the state (say when the thumbnail is loaded), we are issuing a duplicate call to this method. Addition of this flag ensures that this method will only be called once in the lifecycle of this widget.

ua741 (approved):

PR #818, opened 2023-01-25
Improve logging and add null checks at missing places

ua741:


lib/ui/shared_collections_gallery.dart
```diff
@@ -69,8 +70,20 @@ class _SharedCollectionGalleryState extends State<SharedCollectionGallery>
...
-              .getCollectionByID(file.collectionID!)!;
+          if (file.collectionID == null) {
+            _logger.severe("collection id should not be null");
+            continue;
+          }
+          final Collection? c =
+              CollectionsService.instance.getCollectionByID(file.collectionID!);
+          if (c == null) {
+            _logger
+                .severe("shared collection is not cached ${file.collectionID}");
```
ashilkn, line 81: This can be both shared or own collection right? `file` is from `getLatestCollectionFiles()`.

ua741: Right.

ashilkn (approved):

PR #821, opened 2023-01-26
Update delete album copy

ua741:
- Add support for custom bodyWidget in action content
- Update delete album copy

image


mnvr (approved):
🙏

PR #824, opened 2023-01-27
Dialog fixes

ashilkn:
## Description

Cleared up the UI and code related to dialogs.


lib/services/user_service.dart
```diff
@@ -197,21 +197,22 @@ class UserService {
   }
 
   Future<void> logout(BuildContext context) async {
-    final dialog = createProgressDialog(context, "Logging out...");
-    await dialog.show();
     try {
+      await Future.delayed(const Duration(seconds: 3));
```
ua741, line 201: Remove?

lib/services/user_service.dart
```diff
@@ -197,21 +197,22 @@ class UserService {
...
-        await dialog.hide();
         Navigator.of(context).popUntil((route) => route.isFirst);
       } else {
         throw Exception("Log out action failed");
       }
     } catch (e) {
       _logger.severe(e);
-      await dialog.hide();
-      showGenericErrorDialog(context: context);
+      Future.delayed(
```
ua741, line 211: What's the reason for delay? Should we add comment for same?

lib/ui/settings/account_section_widget.dart
```diff
@@ -157,45 +156,13 @@ class AccountSectionWidget extends StatelessWidget {
...
-          },
-        ),
-      ],
-    );
-
-    await showDialog(
-      context: context,
-      builder: (BuildContext context) {
-        return alert;
+    showChoiceActionSheet(
```
ua741, line 159: Add await?

lib/services/user_service.dart
ashilkn: Will add a comment, this future needed so that the first dialog gets dismissed before this comes up.

lib/ui/settings/account_section_widget.dart
ashilkn: Have changed the return type from `Future<void>` to `void`. Don't have to await on it.

ua741 (approved):

ua741 (approved):

PR #832, opened 2023-01-31
Refactor: Clean up local sync service

ua741:
Note: We don't keep a track of edited file IDs anymore. So removed the code related to that.

### DIFF
--- a/lib/services/local_sync_service.dart
+++ b/lib/services/local_sync_service.dart
@@ -109,8 +109,7 @@ class LocalSyncService {
         toTime: syncStartTime,
       );
     }
-    if (!_prefs.containsKey(kHasCompletedFirstImportKey) ||
-        !(_prefs.getBool(kHasCompletedFirstImportKey)!)) {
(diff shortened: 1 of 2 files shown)


lib/services/local_sync_service.dart
```diff
@@ -330,19 +335,15 @@ class LocalSyncService {
...
+          (file) =>
+              file.localID != null &&
+              existingLocalFileIDs.contains(file.localID),
+        )
+        .map((e) => e.localID!)
         .toList();
-    if (updatedFiles.isNotEmpty) {
-      _logger.info(
-        updatedFiles.length.toString() + " local files were updated.",
-      );
```
ashilkn, line 362: Why remove this logging?

ashilkn (approved):

lib/services/local_sync_service.dart
ua741: Enhanced the logging inside `FileUpdationDB.instance.insertMultiple`

PR #847, opened 2023-02-06
feat[gallery_extention]: integrate handler for gallery pick intent

Muhesh7:
## Description

*  Integrated intent action handler for pick intent using `media_extensions` plugin
   - Allows the app to identify and handle case of our app's invocation by other app via what kind of intent-action.
    By invoking getIntentAction method, the following intentActions can be identified.
      * PICK
      * EDIT
      * VIEW
   - Provides setResult method to pass the uri of the selected image to the requesting app by creating a temporary copy of that image in the content-provider path.

## Test Plan

Tested on Android 12: 
* Testing Approach Used:
  * Open WhatsApp Gallery picker option.
  * Choose `ente app` as the gallery picker.
  * Pick an image from it which is sent to WhatsApp as the result.


android/app/build.gradle
```diff
@@ -79,7 +79,7 @@ android {
         fdroid {
             dimension "default"
             applicationIdSuffix ".fdroid"
-            signingConfig null
+            // signingConfig null
```
ua741, line 82: undo?

android/app/src/main/AndroidManifest.xml
```diff
@@ -54,6 +68,15 @@
             </intent-filter>
 
         </activity>
+        <provider
```
ua741, line 71: Can you check if the app works without adding it? Wondering why it was working for set-as option.

android/build.gradle
```diff
@@ -2,6 +2,7 @@ buildscript {
     ext.kotlin_version = '1.6.21'
     repositories {
         google()
+        mavenCentral()  //add this line
```
ua741, line 5: is it needed?

pubspec.yaml
```diff
@@ -77,7 +77,9 @@ dependencies:
   logging: ^1.0.1
   lottie: ^1.2.2
   media_extension:
-    git: "https://github.com/ente-io/media_extension.git"
+    git: 
+      url: "https://github.com/ente-io/media_extension.git"
+      ref: "intent-action"
```
ua741, line 82: can switch to master now :)

ua741 (approved):

ua741 (approved):

PR #852, opened 2023-02-08
Setup TFLite for object detection

vishnukvmd:
## Description
- Setup TFLite for both Android and iOS
- Run a basic model to verify that the integration works as expected without blocking the main thread
- Added a `setup.sh` and updated `README.md`


assets/models/labelmap.txt
```diff
@@ -0,0 +1,91 @@
...
+sheep
+cow
+elephant
+bear
+zebra
+giraffe
+???
+backpack
+umbrella
+???
```
ua741, line 30: ???

pubspec.yaml
```diff
@@ -107,6 +107,11 @@ dependencies:
   step_progress_indicator: ^1.0.2
   syncfusion_flutter_core: ^19.2.49
   syncfusion_flutter_sliders: ^19.2.49
+  tflite_flutter: ^0.9.0
+  tflite_flutter_helper:
```
ua741, line 111: For future, should we document about this custom dependency?

assets/models/labelmap.txt
vishnukvmd: Had imported the original labels as is. Removing `???` seems to not break anything, so have removed.

ua741 (approved):

PR #856, opened 2023-02-10
ButtonWidget : Return exception with buttonAction on dialogs

ashilkn:
## Description

Now dialogs and action sheets which have `ButtonWidget` inside them when popped from the navigator stack will return a new object `ButtonResult` which has properties `ButtonResult.action` and `ButtonResult.exception`. `ButtonResult.exception` will be null if there is no exception on execution of `onTap`.


lib/models/search/button_result.dart
```diff
@@ -0,0 +1,7 @@
+import "package:photos/ui/components/button_widget.dart";
+
+class ButtonResult {
+  final ButtonAction? action;
```
ua741, line 4: Should we add a comment here about when action value can be null?

lib/models/search/button_result.dart
```diff
@@ -0,0 +1,7 @@
...
+  final Exception? exception;
+  ButtonResult({required this.action, this.exception});
```
ua741, line 6: Can reduce verbosity by not using named parameters? Both params are of different types.

ua741 (approved):

ua741 (approved):

Issue #645, opened 2022-11-15
[Feature request] Act as Android's gallery app


tommytran732:
The app currently does not act as Android's gallery app. Tapping on pictures in the File Manager (documentsUI) will not work if no other gallery app is installed.

I would be nice if you can have it acting as a gallery app as well.

mnvr:
Right, this is indeed on our roadmap! https://roadmap.ente.io/set-as-default-gallery-app-p-1528/ 

Thank you for the +1. It is unfortunately not a small undertaking, but we we will try to get to this sooner indeed!

PR #876, opened 2023-02-21
feat(file_viewer): implement file view handler

Muhesh7:
## Description
Implement Image/Video Viewer feature, Through this users can set ente app as their default gallery app and view video/image file from other apps.

Related to #645
## Test Plan
- Tested on Android 12
   - Video File in WhatsApp is opened with ente app and played successfully
   - Image in WhatsApp is opened with ente app and viewed successfully

### DIFF
--- a/lib/app.dart
+++ b/lib/app.dart
@@ -41,7 +41,9 @@ class _EnteAppState extends State<EnteApp> with WidgetsBindingObserver {
   }
 
   void setupIntentAction() async {
-    final mediaExtentionAction = await initIntentAction();
+    final mediaExtentionAction = Platform.isAndroid
+        ? await initIntentAction()
+        : MediaExtentionAction(action: IntentAction.main);
     AppLifecycleService.instance.setMediaExtensionAction(mediaExtentionAction);
     if (mediaExtentionAction.action == IntentAction.main) {
       _configureBackgroundFetch();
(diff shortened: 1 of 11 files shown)


lib/app.dart
```diff
@@ -31,63 +31,44 @@ class EnteApp extends StatefulWidget {
...
-      intentAction = actionResult.action!;
-    } on PlatformException {
-      intentAction = IntentAction.main;
-    } catch (error) {
-      _logger.info(error);
-      intentAction = IntentAction.main;
-    }
-    if (intentAction == IntentAction.main) {
+  void setupIntentAction() async {
+    final mediaExtentionAction = await initIntentAction();
```
ua741, line 44: Let's change this to something like this?


`
final mediaExtentionAction = Platform.isAndroid ? await initIntentAction() : MediaExtentionAction(action: Intent.main)
`

ua741 (dismissed):

lib/app.dart
Muhesh7: But by default the value of AppLifeCycleService.Instance.mediaExtensionAction = MediaExtentionAction(action: IntentAction.main);
https://github.com/ente-io/photos-app/pull/876/files#diff-74b0f1f5e15090884b7d3f16a0b37b04e73ac886e6e616c9cbbbba9bb9815d3bR8

ua741 (approved):

PR #883, opened 2023-02-23
Fix: Rendering of archived albums inside archive section

ua741:
- Fix: Left Align the albums list
- Refactor:create widget for horizontal album listing
- Fix bug in state refresh
- Shrink horizontal album list when empty

### DIFF
--- a/lib/ui/components/album_horizontal_list_widget.dart
+++ b/lib/ui/components/album_horizontal_list_widget.dart
@@ -53,6 +53,9 @@ class _AlbumHorizontalListWidgetState extends State<AlbumHorizontalListWidget> {
           _logger.severe("failed to fetch albums", snapshot.error);
           return const Text("Something went wrong");
         } else if (snapshot.hasData) {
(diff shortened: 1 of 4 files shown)


lib/ui/components/album_horizontal_list_widget.dart
```diff
@@ -9,7 +9,8 @@ import "package:photos/ui/collections/collection_item_widget.dart";
 import "package:photos/ui/common/loading_widget.dart";
 
 class AlbumHorizontalListWidget extends StatefulWidget {
-  final Future<List<CollectionWithThumbnail>> future;
+  final Future<List<CollectionWithThumbnail>> Function() future;
```
ashilkn, line 13: Should we name this `CollectionsWithThumbnail` instead of `future`?

ua741: for brevity, changed it to `collectionsFuture`

ashilkn (approved):

PR #941, opened 2023-03-27
Add location screen

ashilkn:
## Description

**Add location sheet**
- All files with location data is fetched from DB on opening the Add location sheet and ignored files are removed from it.
- Then from the list of files, files that do not come under the location tag are removed.
- The gallery renders upto 1000 files.

https://user-images.githubusercontent.com/77285023/227916698-6863ee00-feb2-4fb5-a51d-8dcbfcd6309d.mp4


lib/services/location_service.dart
```diff
@@ -73,6 +73,21 @@ class LocationService {
     return result;
   }
 
+  isFileInsideLocationTag(
```
vishnukvmd, line 76: Return type `bool`?

lib/ui/huge_listview/lazy_loading_gallery.dart
```diff
@@ -403,12 +406,12 @@ class _LazyLoadingGridViewState extends State<LazyLoadingGridView> {
         mainAxisSpacing: 2,
         crossAxisCount: widget.photoGridSize!,
       ),
-      padding: const EdgeInsets.all(0),
+      padding: const EdgeInsets.symmetric(vertical: (galleryGridSpacing / 2)),
```
vishnukvmd, line 409: Is this the one that was causing missing spaces between certain rows?

lib/ui/viewer/location/add_location_sheet.dart
```diff
@@ -0,0 +1,216 @@
...
+  @override
+  State<AddLocationSheet> createState() => _AddLocationSheetState();
+}
+
+class _AddLocationSheetState extends State<AddLocationSheet> {
+  //When memoriesCountNotifier is null, we show the loading widget in the
+  //memories count section which also means the gallery is loading.
+  ValueNotifier<int?> memoriesCountNotifier = ValueNotifier(null);
+  //The value of this notifier has no significance.
+  ValueNotifier<bool> submitNotifer = ValueNotifier(false);
```
vishnukvmd, line 57: Doubt: when do we prefix with `_` and when do we not?

vishnukvmd (approved):

PR #942, opened 2023-03-27
Location info item in file details

ashilkn:
## Description

- Wrote logic to check whether to show the Location InfoItem by checking if the file has GPS data in its EXIF.
- UI for location InfoItem.
- Get location tags associated with a file and render them as `ChipButtonWidgets`.


lib/services/location_service.dart
```diff
@@ -81,3 +116,21 @@ class LocationService {
     return map;
   }
 }
+
+class GPSData {
+  String latRef;
```
vishnukvmd, line 121: All of these could be final?

lib/ui/viewer/file/file_details_widget.dart
```diff
@@ -200,6 +237,36 @@ class _FileDetailsWidgetState extends State<FileDetailsWidget> {
...
   }
 
+  bool _haGPSData() {
+    return _exifData["lat"] != null &&
+        _exifData["long"] != null &&
+        _exifData["latRef"] != null &&
+        _exifData["longRef"] != null;
+  }
+
+  _generateExifForLocation(Map<String, IfdTag> exif) {
```
vishnukvmd, line 247: Could define the return type as `void`

vishnukvmd (approved):

PR #984, opened 2023-04-13
Fix: Open location screen on locationSearch result

ua741:


lib/models/search/generic_search_result.dart
```diff
@@ -1,12 +1,14 @@
+import "package:flutter/cupertino.dart";
```
ashilkn, line 1: Why was this added?

ashilkn (approved):

lib/models/search/generic_search_result.dart
ua741: context

PR #1003, opened 2023-04-18
Fix: Clear entities table on logout

ua741:
- Fix: clear entities table on logout
- [generated] i10n changes


lib/generated/intl/messages_zh.dart
ua741, line 1: Not sure why this was not collapsed. Checking

ua741: https://github.com/ente-io/photos-app/pull/1004 Will merge this changes in this diff when the other diff is landed in main

vishnukvmd (approved):

PR #1028, opened 2023-04-24
DMS to Decimal Degree(DD) parsing bug fix

ashilkn:
## Description

There was an arithmetic bug in parsing of location data from DMS to DD. This PR fixes this bug.

## Test Plan

Tested with the following input cases : 
- `latRef`, `lat`, `longRef`, `long` are all not `null`
- `latRef` and `longRef` are `null`
- `lat` & `long` are `null`
- All are `null`

Tested some other cases which are very unlikely to occur: 
- `latRef` and `longRef` are `null` and lat and long are signed (`[-24, 41, 23.45]`).
- `latRef` and `longRef` are not `null` and `lat` and `long` are signed.

### DIFF
--- a/lib/services/location_service.dart
+++ b/lib/services/location_service.dart
@@ -215,8 +215,8 @@ class GPSData {
     final latSign = (latRef ?? 'N') == "N" ? 1 : -1;
     final longSign = (longRef ?? 'E') == "E" ? 1 : -1;
     return Location(
-      latitude: latSign * lat![0] + lat![1] / 60 + lat![2] / 3600,
-      longitude: longSign * long![0] + long![1] / 60 + long![2] / 3600,
+      latitude: latSign * (lat![0] + lat![1] / 60 + lat![2] / 3600),
+      longitude: longSign * (long![0] + long![1] / 60 + long![2] / 3600),
     );
   }
 }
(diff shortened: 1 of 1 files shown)


lib/services/location_service.dart
```diff
@@ -206,17 +206,32 @@ class GPSData {
...
+      }
+    } else {
+      latSign = latRef == "N" ? 1 : -1;
+      longSign = longRef == "E" ? 1 : -1;
+    }
+
     return Location(
-      latitude: latSign * lat![0] + lat![1] / 60 + lat![2] / 3600,
-      longitude: longSign * long![0] + long![1] / 60 + long![2] / 3600,
+      latitude: latSign * (lat![0] + lat![1] / 60 + lat![2] / 3600),
```
paulhovey, line 233: I personally like explicit order grouping for all operations:

`latitude: latSign * (lat![0] + (lat![1] / 60) + (lat![2] / 3600))`

Helps eliminate bugs like this one :)

ua741 (approved):

lib/services/location_service.dart
ashilkn: Yes, that's good to have indeed :)

PR #1045, opened 2023-04-29
Tests for exif to location parsing

ashilkn:
## Description

Tests for EXIF to location parsing.

### DIFF
--- a/lib/services/location_service.dart
+++ b/lib/services/location_service.dart
@@ -225,15 +225,14 @@ class GPSData {
         long![long!.indexOf(element)] = element.abs();
       }
     } else {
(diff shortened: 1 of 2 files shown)


lib/services/location_service.dart
```diff
@@ -198,16 +198,16 @@ class LocationService {
 }
 
 class GPSData {
-  final String? latRef;
+  String? latRef;
```
ua741, line 201: Why not keep this as final? I don't think modifying the value when toLocationObject is called is recommended approach. If needed, define additional variables inside `toLocationObject` or create a uil method `converToLocation(GPSData data)`

ua741 (approved):

PR #1053, opened 2023-05-02
Minor bug fixes

ua741:
- Fix bug: send publicMagicMetadata during upload
- Handle infinite or Nan values

### DIFF
--- a/lib/models/location/location.dart
+++ b/lib/models/location/location.dart
@@ -18,10 +18,10 @@ class Location with _$Location {
     if (location.latitude == null || location.longitude == null) return false;
     final latValue = location.latitude!;
     final longValue = location.longitude!;
-    if (latValue.isNaN || latValue.isNaN || latValue == 0.0) {
+    if (latValue.isNaN || latValue.isInfinite || latValue == 0.0) {
       return false;
     }
-    if (longValue.isInfinite || longValue.isInfinite || longValue == 0.0) {
+    if (longValue.isNaN || longValue.isInfinite || longValue == 0.0) {
       return false;
     }
     return true;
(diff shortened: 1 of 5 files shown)


lib/models/location/location.dart
```diff
@@ -12,4 +12,18 @@ class Location with _$Location {
...
+
+  static isValidLocation(Location? location) {
+    if (location == null) return false;
+    if (location.latitude == null || location.longitude == null) return false;
+    final latValue = location.latitude!;
+    final longValue = location.longitude!;
+    if (latValue.isNaN || latValue.isNaN || latValue == 0.0) {
+      return false;
+    }
+    if (longValue.isInfinite || longValue.isInfinite || longValue == 0.0) {
```
ashilkn, line 24: .

ashilkn (approved):

lib/models/location/location.dart
```diff
@@ -12,4 +12,18 @@ class Location with _$Location {
...
+
+  static isValidLocation(Location? location) {
+    if (location == null) return false;
+    if (location.latitude == null || location.longitude == null) return false;
+    final latValue = location.latitude!;
+    final longValue = location.longitude!;
+    if (latValue.isNaN || latValue.isInfinite || latValue == 0.0) {
+      return false;
+    }
+    if (longValue.isNaN || longValue.isInfinite || longValue == 0.0) {
```
paulhovey, line 24: Latitude and Longitude of 0 are valid values.  The Prime Meridian is even a tourist attraction 😄 something like NaN should be used to store invalid values.

ashilkn: Hey, thanks for pointing this out! Have merged a PR to fix this. Now, a location will be considered invalid only if both latitude and longitude are 0. We know that (0,0) is a valid location, but some older Android versions were causing trouble by tagging location as (0,0). We put in a fix and a migration to handle this. For the time being, we're gonna keep treating (0,0) as invalid to avoid any issues.

paulhovey: Ah it's the worst when technically valid values like 0,0 are used as invalid.  Thanks for fixing!  👍

PR #1057, opened 2023-05-03
HotFix Release: Fix invalid location check & file update

ua741:

### DIFF
--- a/lib/services/local_file_update_service.dart
+++ b/lib/services/local_file_update_service.dart
@@ -257,9 +257,7 @@ class LocalFileUpdateService {
 
       for (ente.File file in enteFiles) {
         final Location? location = await tryLocationFromExif(file);
-        if (location != null &&
-            (location.latitude ?? 0) != 0.0 &&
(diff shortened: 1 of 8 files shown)


lib/services/local_file_update_service.dart
```diff
@@ -257,9 +257,7 @@ class LocalFileUpdateService {
 
       for (ente.File file in enteFiles) {
         final Location? location = await tryLocationFromExif(file);
-        if (location != null &&
-            (location.latitude ?? 0) != 0.0 &&
-            (location.longitude ?? 0) != 0.0) {
+        if (location != null && Location.isValidLocation(location)) {
```
ashilkn, line 260: Can remove the `location != null` check as `Location.isValidLocation()` checks that and returns false.

ua741: So added this because idea was crying while we were using it in the bottom.

ashilkn (approved):

vishnukvmd (approved):

ua741:
Closing this. Latest release has these fixes

PR #1097, opened 2023-05-15
Perf improvements

ashilkn:
## Description

- Decode image for thumbnail that have hight and width in public metadata at the most minimum dimensions to reduce the work done on the raster thread. 
- Image will be decoded at the a size where it's smallest side will be equal to the size of a child in the grid.
- This optimisation has reduced the 99th percentile frame rasterizer time, which are frames which cause jank. 
- The work on the UI thread has increased a bit, but is within limits. 
- Used integration tests + flutter driver to get performance data on scrolling gallery.


lib/ui/viewer/file/thumbnail_widget.dart
```diff
@@ -58,9 +58,12 @@ class _ThumbnailWidgetState extends State<ThumbnailWidget> {
   bool _isLoadingRemoteThumbnail = false;
   bool _errorLoadingRemoteThumbnail = false;
   ImageProvider? _imageProvider;
+  int? optimizedImageHeight;
+  int? optimizedImageWidth;
 
   @override
   void initState() {
+    assignOptimizedImageDimensions();
```
ua741, line 66: super.initiate(); should be first line

lib/ui/account/password_reentry_page.dart
```diff
@@ -169,6 +170,7 @@ class _PasswordReentryPageState extends State<PasswordReentryPage> {
                 Padding(
                   padding: const EdgeInsets.fromLTRB(20, 24, 20, 0),
                   child: TextFormField(
+                    key: const ValueKey("passwordInputField"),
```
ua741, line 173: what's the benefit of adding these valueKeys. By default, it seems to be missing in most tutorials/examples.

ua741 (approved):

pubspec.lock
```diff
@@ -53,10 +53,10 @@ packages:
     dependency: "direct main"
     description:
       name: archive
-      sha256: d6347d54a2d8028e0437e3c099f66fdb8ae02c4720c1e7534c9f24c10351f85d
+      sha256: "80e5141fafcb3361653ce308776cfd7d45e6e9fbb429e14eec571382c0c5fecb"
       url: "https://pub.dev"
     source: hosted
-    version: "3.3.6"
+    version: "3.3.2"
```
ua741, line 59: Can we keep the version same>

ua741 (approved):

ua741 (approved):

PR #1157, opened 2023-05-26
Force social links to be opened in the browser instead of the webview

vishnukvmd:


ua741:
Let's only do this for discord in droid? Taking user outside of the application for all links might be overkill?

vishnukvmd:
> Let's only do this for discord in droid? Taking user outside of the application for all links might be overkill?

Ideally we shouldn't be opening any of these links in the web view, since
1. Chances of the user having an active session on any of these portals is higher within their browser
2. Links like Blog and Shop are better experienced on the browser

ua741 (approved):

PR #1158, opened 2023-05-26
Collages

vishnukvmd:
# Collages

- [x] Ability to select between 2-6 photos and create a collage
- [x] Ability to save the created collage to local gallery


lib/models/selected_files.dart
```diff
@@ -5,7 +5,7 @@ import 'package:photos/events/clear_selections_event.dart';
 import 'package:photos/models/file.dart';
 
 class SelectedFiles extends ChangeNotifier {
-  final files = <File>{};
+  final files = <File>[];
```
ua741, line 8: Changing this will cause performance degradation while selecting/updating the states while selecting large number of elements. I don't know if we should make this change just for collage (which will be active for just 6 selections)

lib/services/collections_service.dart
```diff
@@ -792,7 +792,8 @@ class CollectionsService {
         collections.add(collection);
       }
       return collections;
-    } catch (e) {
+    } catch (e, s) {
+      _logger.severe(e, s);
```
ua741, line 796: Why log network errors on sentry? Is this log really required?

lib/ui/tools/collage/collage_creator_page.dart
```diff
@@ -0,0 +1,88 @@
...
+  final _logger = Logger("CreateCollagePage");
+
+  final List<File> files;
+
+  CollageCreatorPage(this.files, {super.key});
+
+  @override
+  Widget build(BuildContext context) {
+    for (final file in files) {
+      _logger.info(file.displayName);
```
ua741, line 25: Can be removed ?

lib/ui/viewer/file_details/upload_icon_widget.dart
```diff
@@ -83,11 +85,18 @@ class _UpdateIconWidgetState extends State<UploadIconWidget> {
           final bool isIgnored = snapshot.data!;
           final bool isQueuedForUpload =
               !isIgnored && widget.file.collectionID != null;
+          if (isQueuedForUpload) {
+            return Padding(
+              padding: const EdgeInsets.all(2.0),
+              child: CircularProgressIndicator(
```
ua741, line 91: This will keep showing up when to there's no network/upload is not in progress or user is out of storage/active sub.

ua741 (approved):

PR #1172, opened 2023-05-31
Deduplication: Refactor + Dedupe by hash and size (instead of just size)

ua741:
## Description
* Defaults to group by size and hash
* Fix state refresh bug for checkbox to reflect correct grouping behaviour
* Refactor the logic for group by filename

### DIFF
--- a/lib/services/deduplication_service.dart
+++ b/lib/services/deduplication_service.dart
@@ -103,25 +103,23 @@ class DeduplicationService {
     return result;
   }
 
-  List<DuplicateFiles> clubDuplicatesByName(List<DuplicateFiles> dupes) {
-    final result = <DuplicateFiles>[];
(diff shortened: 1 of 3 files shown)


lib/services/deduplication_service.dart
```diff
@@ -103,42 +103,56 @@ class DeduplicationService {
...
         }
       }
-      files.removeWhere((file) => incorrectDuplicates.contains(file));
-      if (files.length > 1) {
-        result.add(DuplicateFiles(files, dupe.size));
+    }
+    return dupesBySizeAndName;
+  }
+
+  List<DuplicateFiles> clubDuplicatesByHash(List<DuplicateFiles> dupesBySize) {
```
ua741, line 132: Can potentially refactor this further to reduce the code. Looking into it.

PR #1174, opened 2023-05-31
Refactor Gallery: No change in logic

ua741:


lib/ui/viewer/gallery/component/huge_gallery_list_view.dart
```diff
@@ -8,12 +8,20 @@ import "package:photos/models/file.dart";
...
-class GalleryListView extends StatelessWidget {
+/*
+HugeGalleryListView is a widget that displays a list of grouped/collated
+files when grouping is enabled.
+For each group, it displays a header and use LazyGroupGallery to display a
+particular group of files.
+If a group has more than 400 files, LazyGroupGallery internally divides the
+group into multiple grid views during rendering.
+ */
+class HugeGalleryListView extends StatelessWidget {
```
ua741, line 24: I couldn't come up with better name for this..
`MultipleGroupsGalleryView` `GalleryGroupedListsView` might be better candidate

lib/ui/viewer/gallery/component/group/group_gallery.dart
```diff
@@ -0,0 +1,61 @@
...
+          selectedFiles,
+          index == 0,
+          files.length > kRecycleLimit,
+          toggleSelectAllFromDay,
+          areAllFromDaySelected,
+          photoGridSize,
+          limitSelectionToOne: limitSelectionToOne,
+        ),
+      );
+    }
```
ua741, line 55: This should probably be a ListView.. Not sure if the scrollbar will work if we change it to that. cc @ashilkn

ashilkn: I understood why. Looking into it.

ashilkn (approved):

PR #1192, opened 2023-06-09
Decrypt fileKey in background worker when decrypting thumbnail

ashilkn:
## Test results

Screenshot 2023-06-09 at 11 13 32 AM

### DIFF
--- a/lib/utils/file_download_util.dart
+++ b/lib/utils/file_download_util.dart
@@ -1,6 +1,7 @@
 import 'dart:io' as io;
 import 'dart:typed_data';
 
+import "package:computer/computer.dart";
 import 'package:dio/dio.dart';
 import 'package:logging/logging.dart';
 import 'package:photos/core/configuration.dart';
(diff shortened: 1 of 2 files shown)


lib/utils/file_download_util.dart
```diff
@@ -70,3 +71,26 @@ Uint8List getFileKey(ente.File file) {
       CollectionsService.instance.getCollectionKey(file.collectionID!);
   return CryptoUtil.decryptSync(encryptedKey, collectionKey, nonce);
 }
+
+Future<Uint8List> getFileKeyUsingBgWorker(ente.File file) async {
```
ua741, line 75: getFileKey also maintains a in-memory cache (as fileKey doesn't change that). We should first check the cache and use the value if present.

ua741 (approved):

lib/utils/file_download_util.dart
ua741: Seems like I was wrong. We don't maintain a cache for fileKeys. It's not super critical to have done.

PR #1224, opened 2023-06-19
Remove dependency from getCollectionsWithThumbnails during collectionDelete

ua741:
- use collectionFileCount for identifying emptyCollection
- StopWatch: Log times during debug mode only
- Fix lint
- Log time metrics for home_page_query
- Refactor delete emptyAlbums logic

### DIFF
--- a/lib/db/files_db.dart
+++ b/lib/db/files_db.dart
@@ -1255,6 +1255,28 @@ class FilesDB {
     return result;
   }
 
+  // getCollectionLatestFileTime returns map of collectionID to the max
+  // creationTime of the files in the collection.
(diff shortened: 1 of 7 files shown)


lib/db/files_db.dart
```diff
@@ -1254,6 +1255,28 @@ class FilesDB {
     return result;
   }
 
+  // getCollectionLatestFileTime returns map of collectionID to the max
+  // creationTime of the files in the collection.
+  Future<Map<int, int>> getCollectionIDToMaxCreationTime() async {
```
ua741, line 1260: We will also need to cache this result to provide sort by newest photo functionality. It's still much faster than `getLatestCollectionFiles`

ashilkn (approved):

PR #1237, opened 2023-06-25
Notifications for shared items

vishnukvmd:
## Description
- [x] Display a notification when the app is in background, and a photo was added to a collection by someone else


lib/main.dart
```diff
@@ -53,7 +52,8 @@ bool _isProcessRunning = false;
 const kLastBGTaskHeartBeatTime = "bg_task_hb_time";
 const kLastFGTaskHeartBeatTime = "fg_task_hb_time";
 const kHeartBeatFrequency = Duration(seconds: 1);
-const kFGSyncFrequency = Duration(minutes: 5);
+const kFGSyncFrequency =
```
ua741, line 55: Can we undo this to avoid noise during normal development.

vishnukvmd: Fixed.

ua741 (approved):

PR #1282, opened 2023-07-17
Minor improvement

ashilkn:
## Description

Store set of files in state of widget rather than converting to set from list in multiple places. Convert and initialise `Set<File>` from `List<File>` only when `Set<File>` is used.


lib/ui/viewer/gallery/component/group/lazy_group_gallery.dart
```diff
@@ -75,6 +76,7 @@ class _LazyGroupGalleryState extends State<LazyGroupGallery> {
     _logger = Logger("LazyLoading_${widget.logTag}");
     _shouldRender = true;
     _files = widget.files;
+    _filesAsSet = _files.toSet();
```
ua741, line 79: Instead of conversion during selection, this will do it during load. May be we can make it a lazy getter

ua741 (approved):

PR #1295, opened 2023-08-02
Remove password auto fill on create new account screen

ashilkn:


ua741:
This is required otherwise OS won't prompt to save.

PR #1343, opened 2023-08-24
Club mutiple defaultHiddenCollections to one

ashilkn:
## Description

- Move files of all `defaultHiddenCollection`s to one `defaultHiddenCollection`
- Delete the excess empty `defaultHiddenCollections`
- If moving files fails, the excess collection won't be trashed. 

## Test Plan

- Tested deletion of multiple excess `defaultHiddenFiles`.

### DIFF
--- a/lib/services/hidden_service.dart
+++ b/lib/services/hidden_service.dart
@@ -72,7 +72,7 @@ extension HiddenService on CollectionsService {
         ))
             .files;
         await move(result.id, defaultHidden.id, filesInCollection);
-        CollectionsService.instance.trashEmptyCollection(defaultHidden);
+        await CollectionsService.instance.trashEmptyCollection(defaultHidden);
       } on AssertionError catch (e) {
         _logger.severe("Clubbing all defaultHidden : ${e.message}");
         continue;
--- a/lib/services/hidden_service.dart
+++ b/lib/services/hidden_service.dart
@@ -73,9 +73,6 @@ extension HiddenService on CollectionsService {
             .files;
         await move(result.id, defaultHidden.id, filesInCollection);
         await CollectionsService.instance.trashEmptyCollection(defaultHidden);
-      } on AssertionError catch (e) {
-        _logger.severe("Clubbing all defaultHidden : ${e.message}");
-        continue;
       } catch (e, s) {
         _logger.severe(
           "One iteration of clubbing all default hidden failed",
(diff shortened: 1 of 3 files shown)


lib/services/hidden_service.dart
```diff
@@ -23,24 +25,70 @@ extension HiddenService on CollectionsService {
...
   }
 
+  Future<Collection> clubAllDefaultHiddenToOne(
+    List<Collection> allDefaultHidden,
+  ) async {
+    final Collection result = allDefaultHidden.first;
+
+    for (Collection defaultHidden in allDefaultHidden) {
+      try {
+        if (defaultHidden == allDefaultHidden.first) {
```
ua741, line 65: let's check for collectionID here.

lib/services/hidden_service.dart
```diff
@@ -23,24 +25,70 @@ extension HiddenService on CollectionsService {
...
+          continue;
+        }
+        final filesInCollection = (await FilesDB.instance.getFilesInCollection(
+          defaultHidden.id,
+          galleryLoadStartTime,
+          galleryLoadEndTime,
+        ))
+            .files;
+        await move(result.id, defaultHidden.id, filesInCollection);
+        CollectionsService.instance.trashEmptyCollection(defaultHidden);
```
ua741, line 75: await on trashEmptyCollection?

lib/services/hidden_service.dart
```diff
@@ -23,24 +25,70 @@ extension HiddenService on CollectionsService {
...
+      } on AssertionError catch (e) {
+        _logger.severe("Clubbing all defaultHidden : ${e.message}");
```
ua741, line 77: why is there special handling for AssertionError?

ashilkn: Thought logging would be more clear in case something goes wrong. Realised that it doesn't make a difference. Have removed it.

lib/services/hidden_service.dart
```diff
@@ -20,27 +22,68 @@ import 'package:photos/utils/dialog_util.dart';
...
   }
 
+  Future<Collection> clubAllDefaultHiddenToOne(
+    List<Collection> allDefaultHidden,
+  ) async {
+    final Collection result = allDefaultHidden.first;
+
+    for (Collection defaultHidden in allDefaultHidden) {
+      try {
+        if (defaultHidden.id == allDefaultHidden.first.id) {
```
ua741, line 63: `defaultHidden.id === result.id`

ua741 (dismissed):

ua741 (approved):

PR #1385, opened 2023-09-09
Clean up state files in temp directory when all uploads are done.

ua741:

### DIFF
--- a/lib/services/remote_sync_service.dart
+++ b/lib/services/remote_sync_service.dart
@@ -139,6 +139,11 @@ class RemoteSyncService {
           Bus.instance.fire(SyncStatusUpdate(SyncStatus.completedBackup));
         }
       } else {
+        // if no files are uploaded and filesToBeUploaded is empty, clear
+        // stale files in the temp directory
+        if(!hasUploadedFiles && filesToBeUploaded.isEmpty) {
+          await _uploader.removeStaleFiles();
+        }
         _existingSync?.complete();
         _existingSync = null;
       }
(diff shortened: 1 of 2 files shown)


lib/utils/file_uploader.dart
```diff
@@ -282,6 +283,29 @@ class FileUploader {
     }
   }
 
+
+  Future<void> removeStaleFiles() async {
+    try {
+      final String dir = Configuration.instance.getTempDirectory();
```
vishnukvmd, line 289: Would it make more sense to create a sub-dir for keeping these files and just removing that?

lib/services/remote_sync_service.dart
```diff
@@ -139,6 +139,11 @@ class RemoteSyncService {
           Bus.instance.fire(SyncStatusUpdate(SyncStatus.completedBackup));
         }
       } else {
+        // if no files are uploaded and filesToBeUploaded is empty, clear
+        // stale files in the temp directory
+        if(!hasUploadedFiles && filesToBeUploaded.isEmpty) {
```
vishnukvmd, line 144: nit: lint

fwiw, I have auto-format configured to run on on file-save

lib/utils/file_uploader.dart
ua741: I initially thought of starting with that but the current configuration file is little too much and confusing and didn't want to refactor that yet. We will fix these when we pick resumable upload and download

lib/services/remote_sync_service.dart
```diff
@@ -139,6 +139,11 @@ class RemoteSyncService {
           Bus.instance.fire(SyncStatusUpdate(SyncStatus.completedBackup));
         }
       } else {
+        // if no files are uploaded and filesToBeUploaded is empty, clear
+        // stale files in the temp directory
+        if (!hasUploadedFiles && filesToBeUploaded.isEmpty) {
```
vishnukvmd, line 144: Could just be `filesToBeUploaded.isEmpty`?

vishnukvmd (approved):

PR #1405, opened 2023-09-18
Remove redundant dependencies

vishnukvmd:


ua741:
Disabled auto-merge. 
I am assuming that we don't need to depend on custom fork now. Don't have full context why it was added in the start. Feel free to merge.

ua741 (approved):

vishnukvmd:
@ua741 Had cloned `extended_image` because it had some changes that later got merged upstream. `flutter_image_compress` had some issues with building on iOS, that was also patched later.

PR #1410, opened 2023-09-20
feat(screen-refresh-rate): use flutter_displaymode to enable high refresh rates in devices with refresh rate locked at 60FPS

ashilkn:


lib/main.dart
```diff
@@ -65,6 +66,7 @@ void main() async {
   final savedThemeMode = await AdaptiveTheme.getThemeMode();
   await _runInForeground(savedThemeMode);
   BackgroundFetch.registerHeadlessTask(_headlessTaskHandler);
+  await FlutterDisplayMode.setHighRefreshRate();
```
ua741, line 69: Do we have to wait on this?

ua741 (approved):

lib/main.dart
ashilkn: Have removed it

ua741 (approved):

PR #1447, opened 2023-10-22
Warn auth and photos accounts are the same

ashilkn:
## Description

- Added a Notification widget on top + Modified confirm deletion prompt of checkbox.
- Layout changes

Before: 

https://github.com/ente-io/photos-app/assets/77285023/944cffb0-71f0-4d8a-bbc8-fc8222c4d832

After:

https://github.com/ente-io/photos-app/assets/77285023/483f130f-3699-4ffd-a9ce-8b8c43381ea4

### DIFF
--- a/lib/generated/intl/messages_de.dart
+++ b/lib/generated/intl/messages_de.dart
@@ -428,6 +428,8 @@ class MessageLookup extends MessageLookupByLibrary {
             "Bist du sicher, dass du die Zwei-Faktor-Authentifizierung (2FA) deaktivieren willst?"),
         "confirmAccountDeletion":
             MessageLookupByLibrary.simpleMessage("Kontolöschung bestätigen"),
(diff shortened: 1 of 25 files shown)


vishnukvmd:
Let's selectively show this only in cases where a user has used both the products? Majority of Photos users might be clueless about Auth, and vice-versa.

ua741:
> Let's selectively show this only in cases where a user has used both the products? Majority of Photos users might be clueless about Auth, and vice-versa.

Do we also want to handle for locker? May be we can have a generic copy

ashilkn:
@ua741 
>>> May be we can have a generic copy

Reverted all the copy changes to the delete account screen (kept the minor layout changes), and changed the copy of the very last deletion confirmation dialog. Can use this across all apps.

Before: 

6815D5B7-FAD7-4F4F-AB31-637BBCDCA59B_1_201_a

After: 

3FD3AA34-7654-4A11-A6DE-56932609A6E9_1_201_a

PR #1457, opened 2023-10-27
Clip

vishnukvmd:


lib/utils/thumbnail_util.dart
```diff
@@ -42,6 +42,18 @@ Future<Uint8List?> getThumbnail(EnteFile file) async {
...
 
+Future<File?> getThumbnailFile(EnteFile file) async {
+  final thumbnail = await getThumbnail(file);
+  if (thumbnail != null) {
+    final cachedThumbnail = cachedThumbnailPath(file);
+    if (!await cachedThumbnail.exists()) {
+      await cachedThumbnail.writeAsBytes(thumbnail, flush: true);
+    }
+    return cachedThumbnail;
+  }
```
ua741, line 53: `cachedThumbnailPath` will throw error if this is called for a file that's not uploaded.
May be rename this to `getRemoteThubnailFile()`?

lib/utils/thumbnail_util.dart
```diff
@@ -42,6 +42,18 @@ Future<Uint8List?> getThumbnail(EnteFile file) async {
   }
 }
 
+Future<File?> getThumbnailFile(EnteFile file) async {
+  final thumbnail = await getThumbnail(file);
+  if (thumbnail != null) {
+    final cachedThumbnail = cachedThumbnailPath(file);
+    if (!await cachedThumbnail.exists()) {
+      await cachedThumbnail.writeAsBytes(thumbnail, flush: true);
+    }
```
ua741, line 51: If we just need the path, it might be better to move this logic above.
Assuming we are going to rename the method to `getRemoteThumbnailFile()`
```
final cachedThumbnail = cachedThumbnailPath(file);
if (await cachedThumbnail.exists()) {
        return cachedThumbnail;
}
final thumbnail = await getThumbnail(file)
if(thumbnail != null) {
 // it might be already written to this path due // `getThumbnail(file)`
  if (!await cachedThumbnail.exists()) {
      await cachedThumbnail.writeAsBytes(thumbnail, flush: true);
    }
     return cachedThumbnail;
}
...
```

lib/services/semantic_search/semantic_search_service.dart
```diff
@@ -0,0 +1,276 @@
...
+    final queryResults = <QueryResult>[];
+    for (final embedding in embeddings) {
+      final score = computeScore({
+        "imageEmbedding": embedding.embedding,
+        "textEmbedding": textEmbedding,
+      });
+      queryResults.add(QueryResult(embedding.fileID, score));
+    }
+    queryResults.sort((first, second) => second.score.compareTo(first.score));
+    queryResults.removeWhere((element) => element.score < kScoreThreshold);
```
ua741, line 128: instead of removeWhere, we can avoid adding such results

ua741 (approved):

ua741 (approved):

PR #1501, opened 2023-11-15
Zoom image sliding

### LINKED ISSUE #1451: Switch to the next photo while panning at the edge of the current photo
If we've already reached the edge of the current photo in the zoomed out state, panning further should switch over to the next photo.

Of the top of my head, we could attach a different `GestureDetector` and use the `scale` attribute of the `PhotoView` to compute this:

https://github.com/ente-io/photos-app/blob/177b1bf3bb42589ccc0a73183a648b3f4023c8d6/lib/ui/viewer/file/zoomable_image.dart#L117

i-aiymen:
## Description

Functionality to switch the image to left or right, even if the images is zoomed

Closes #1451

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -54,14 +54,14 @@ android {
         multiDexEnabled true
     }
 
-    signingConfigs {
-       release {
-           storeFile keystoreProperties['storeFile'] ? file(keystoreProperties['storeFile']) : file(System.getenv("SIGNING_KEY_PATH"))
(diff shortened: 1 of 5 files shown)


i-aiymen:
@vishnukvmd

vishnukvmd:
Hey, this change breaks the ability to pan through a zoomed image. What we want is the ability to pan, until the edge of the image, and swipe to the next one once we've hit the edge.

PR #1502, opened 2023-11-16
Fixed: When we zoom the image, we will be able to slide to next or pr…

### LINKED ISSUE #1451: Switch to the next photo while panning at the edge of the current photo

i-aiymen:
## Description

Functionality to switch the image to left or right, even if the images is zoomed

Closes #1451

### DIFF
--- a/lib/ui/viewer/file/zoomable_image.dart
+++ b/lib/ui/viewer/file/zoomable_image.dart
@@ -6,6 +6,7 @@ import 'package:flutter/material.dart';
 import 'package:flutter/widgets.dart';
 import 'package:logging/logging.dart';
 import 'package:photo_view/photo_view.dart';
+import "package:photo_view/photo_view_gallery.dart";
 import 'package:photos/core/cache/thumbnail_in_memory_cache.dart';
 import 'package:photos/core/constants.dart';
 import 'package:photos/core/event_bus.dart';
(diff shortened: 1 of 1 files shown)


i-aiymen:
@vishnukvmd Just check it out

lib/ui/viewer/file/zoomable_image.dart
```diff
@@ -89,21 +90,17 @@ class _ZoomableImageState extends State<ZoomableImage>
...
-              ? PhotoViewComputedScale.covered
-              : PhotoViewComputedScale.contained,
-          gaplessPlayback: true,
-          heroAttributes: PhotoViewHeroAttributes(
-            tag: widget.tagPrefix! + _photo.tag,
-          ),
-          backgroundDecoration: widget.backgroundDecoration as BoxDecoration?,
-        ),
+      content = PhotoViewGallery.builder(
+        builder: (context, index) {
```
vishnukvmd, line 94: All of the missing parameters (axis, controller, scaleStateChangedCallback, ...) are necessary.

i-aiymen: Actually, it is needed for photoView widget, for PhotoViewGallery.builder, we don't need all that. I have tested by adding that and removing it. It is of no difference @vishnukvmd

vishnukvmd: All of these parameters are either used upstream, when they're injected from a widget above in the render tree (for eg. `minScale` and `backgroundDecoration`), or used within this same widget (for eg. `gaplessPlayback` and `controller` ensures that the thumbnail is replaced by the full-res image seamlessly).

Looking into their usages will give you a better idea.

i-aiymen: ok

i-aiymen:
@vishnukvmd Updated

vishnukvmd (approved):

PR #1523, opened 2023-11-20
Zoom image sliding bug

i-aiymen:
… callback parameter used inside photo view hallery widget

## Description
Bug resolved: Zoom image sliding being prevented because of state change callback parameter used inside photo view gallery widget


i-aiymen:
@vishnukvmd just check this out, when I just tried to slide when zoomed, it was again restricted. now its resolved

lib/ui/viewer/file/zoomable_image.dart
```diff
@@ -54,23 +54,13 @@ class _ZoomableImageState extends State<ZoomableImage>
...
   PhotoViewController _photoViewController = PhotoViewController();
 
   @override
   void initState() {
     _photo = widget.photo;
     _logger = Logger("ZoomableImage");
     _logger.info('initState for ${_photo.generatedID} with tag ${_photo.tag}');
-    _scaleStateChangedCallback = (value) {
-      if (widget.shouldDisableScroll != null) {
-        widget.shouldDisableScroll!(value != PhotoViewScaleState.initial);
```
vishnukvmd, line 68: Do we then need `shouldDisableScroll`?

lib/ui/viewer/file/zoomable_image.dart
```diff
@@ -111,21 +101,16 @@ class _ZoomableImageState extends State<ZoomableImage>
...
-                      {Navigator.of(context).pop()},
-                    }
-                  else if (d.delta.dy < (dragSensitivity * -1))
-                    {
-                      showDetailsSheet(context, widget.photo),
-                    },
-                },
-            };
+    verticalDragCallback(d) => {
+          if (d.delta.dy > dragSensitivity)
```
vishnukvmd, line 105: The expectation is that if the photo is zoomed in, swiping down should NOT close the photo, and swiping up should NOT bring up the details-sheet. Could you confirm that after removing the `!_isZooming` check this behavior is intact?

i-aiymen: it is having no effect, I have checked it. Even after removing and zoomed and when we swipe down or up, same effect

lib/ui/viewer/file/zoomable_image.dart
i-aiymen: no actually

i-aiymen:
@vishnukvmd Updated

lib/ui/viewer/file/zoomable_image.dart
```diff
@@ -54,23 +54,15 @@ class _ZoomableImageState extends State<ZoomableImage>
   bool _loadedLargeThumbnail = false;
   bool _loadingFinalImage = false;
   bool _loadedFinalImage = false;
-  ValueChanged<PhotoViewScaleState>? _scaleStateChangedCallback;
-  bool _isZooming = false;
   PhotoViewController _photoViewController = PhotoViewController();
+  double? initialScale;
+  double percentage = 42.5 / 100;
```
vishnukvmd, line 59: What's this [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming))? 👀

lib/ui/viewer/file/zoomable_image.dart
vishnukvmd: Then shouldn't we remove it from the Widget as well? And from whoever is passing it down here?

lib/ui/viewer/file/zoomable_image.dart
```diff
@@ -292,6 +281,9 @@ class _ZoomableImageState extends State<ZoomableImage>
         initialPosition: newPosition,
         initialScale: scale,
       );
+      setState(() {
```
vishnukvmd, line 284: This method (`_updatePhotoViewController`) is only called after the final image has been loaded. In case of images that don't exist locally, this could take a while. So the `initialScale` will be `null` until then, affecting this whole behavior

lib/ui/viewer/file/zoomable_image.dart
i-aiymen: It's just percentage value to detect, if the image is zoomed or not from its initial scale value, because for each image, we are calculating the scale by some calculations and then assigning the initial scale value. And if that scale value is less than  the current scale value, then the image is zoomed. So just added a percentage value to check the increase value.

But just now, got another solution. Will update that soon

lib/ui/viewer/file/zoomable_image.dart
i-aiymen: oh I see

i-aiymen: So, while loading, what will be shown then? @vishnukvmd

lib/ui/home/memories/full_screen_memory.dart
```diff
@@ -327,14 +327,6 @@ class _FullScreenMemoryState extends State<FullScreenMemory> {
             file,
             autoPlay: false,
             tagPrefix: "memories",
-            shouldDisableScroll: (value) {
```
vishnukvmd, line 330: Is `_shouldDisableScroll` needed then?

lib/ui/viewer/file/zoomable_image.dart
```diff
@@ -112,21 +100,19 @@ class _ZoomableImageState extends State<ZoomableImage>
       content = const EnteLoadingWidget();
     }
 
-    final GestureDragUpdateCallback? verticalDragCallback = _isZooming
-        ? null
-        : (d) => {
-              if (!_isZooming)
+    verticalDragCallback(d) => {
+          if (_photoViewController.scale! <= _initialScale!)
```
vishnukvmd, line 104: Can you not compare this with `PhotoViewScaleState.initial` like how it was being done previously?

android/app/build.gradle
```diff
@@ -54,14 +54,14 @@ android {
...
 
-    signingConfigs {
-       release {
-           storeFile keystoreProperties['storeFile'] ? file(keystoreProperties['storeFile']) : file(System.getenv("SIGNING_KEY_PATH"))
-           keyAlias keystoreProperties['keyAlias'] ? keystoreProperties['keyAlias'] : System.getenv("SIGNING_KEY_ALIAS")
-           keyPassword keystoreProperties['keyPassword'] ? keystoreProperties['keyPassword'] : System.getenv("SIGNING_KEY_PASSWORD")
-           storePassword keystoreProperties['storePassword'] ? keystoreProperties['storePassword'] : System.getenv("SIGNING_STORE_PASSWORD")
-       }
-    }
+    // signingConfigs {
```
vishnukvmd, line 57: You should be able to use `independentDebug` without any signing configs

vishnukvmd: Also, revert this?

lib/ui/viewer/file/zoomable_image.dart
vishnukvmd: Ref. `PhotoViewScaleState.initial`

lib/ui/home/memories/full_screen_memory.dart
i-aiymen: No needed

lib/ui/viewer/file/zoomable_image.dart
i-aiymen: for that, we need _scaleStateChangedCallback, which then breaks the sliding

i-aiymen:
@vishnukvmd Updated

lib/ui/viewer/file/zoomable_image.dart
i-aiymen: initialScale null case is also updated

lib/ui/viewer/file/zoomable_image.dart
```diff
@@ -112,21 +100,26 @@ class _ZoomableImageState extends State<ZoomableImage>
...
+              {Navigator.of(context).pop()},
+            }
+          else if (d.delta.dy < (dragSensitivity * -1))
+            {
+              showDetailsSheet(context, widget.photo),
+            },
+        };
+
+    verticalDragCallback(d) {
+      if (_initialScale != null &&
```
vishnukvmd, line 115: Can be simplified to 
```dart
if (_initialScale == null || _photoViewController.scale! <= _initialScale!) {
    dragFunction(d);
}
```

i-aiymen:
@vishnukvmd Updated

vishnukvmd (approved):

PR #1556, opened 2023-12-01
LocalFileUpdater: Process live photos

ua741:
## Test Plan
Tested locally

### DIFF
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -503,11 +503,12 @@ packages:
   file_saver:
     dependency: "direct main"
     description:
-      name: file_saver
-      sha256: "591d25e750e3a4b654f7b0293abc2ed857242f82ca7334051b2a8ceeb369dac8"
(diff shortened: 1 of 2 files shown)


lib/db/files_db.dart
```diff
@@ -1449,6 +1449,28 @@ class FilesDB {
...
+  Future<List<String>> getLivePhotosWithBadSize(
+    int userId,
+    int sizeInBytes,
+  ) async {
+    final db = await instance.database;
+    final rows = await db.query(
+      filesTable,
+      columns: [columnLocalID],
+      distinct: true,
+      where: '($columnOwnerID = ? AND $columnOwnerID IS NOT NULL) AND '
```
vishnukvmd, line 1462: Do we need to check for `$columnOwnerID IS NOT NULL`?

vishnukvmd (approved):

PR #1583, opened 2023-12-11
Add option to remove files from uncategorized if they are present in other albums owned by user.

### LINKED ISSUE #1540: Adding items from Uncategorized should remove them from Uncategorized
Should change the implementation to reflect this more like a "Move" than an "Add".

Alternately, provide a way within Uncategorized to clean up items that are already present in other albums.

The latter might be simpler to implement.

ua741:
## Description
Need to decide on icon and label for the option.
image

Closes #1540
## Test Plan
Tested locally.

### DIFF
--- a/lib/ui/actions/collection/collection_sharing_actions.dart
+++ b/lib/ui/actions/collection/collection_sharing_actions.dart
@@ -1,3 +1,5 @@
+import "dart:async";
+
 import 'package:flutter/material.dart';
 import 'package:logging/logging.dart';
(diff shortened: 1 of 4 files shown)


ashilkn (approved):

ashilkn:
I feel the exiting icon is good enough, the only alternative I found was this. 

Screenshot 2023-12-11 at 9 42 19 PM

There is a brush icon, but it didn't look good for some reason compared to these.

PR #1586, opened 2023-12-13
Revert archiving incoming shared in settings

ashilkn:
## Description

Reverting this as to archive photos and videos of an incoming shared album from the home gallery, the album can be archived.


ua741:
We can just revert the merge commit?
git revert -m 1 <mergeCommitID>

PR #1593, opened 2023-12-16
Add/Edit location data feature

### LINKED ISSUE #1441: Ability to set GPS coordinates for photos without location
Since we already have the map view, we could re-use that to let users pick a point on the map.

ashilkn:
## Description

Support for adding or editing location data of files.

closes #1441

https://github.com/ente-io/photos-app/assets/77285023/c4be37cf-e34b-4254-81da-959f6db3c4aa

## Test Plan

Did basic testing, need to do internal testing.

### DIFF
--- a/lib/generated/l10n.dart
+++ b/lib/generated/l10n.dart
@@ -8197,6 +8197,16 @@ class S {
       args: [],
     );
   }
+
+  /// `Edits to location will only be seen within Ente`
+  String get editsToLocationWillOnlyBeSeenWithinEnte {
+    return Intl.message(
+      'Edits to location will only be seen within Ente',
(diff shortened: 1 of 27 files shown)


ua741:
In the dialog, should be mention that only location metadata will be added, and original file won't be modified.

We can may be have an info icon that can show this in a dialog. Can also mention that this location information will be part of extracted metadata (will need to work on the copy)

ashilkn:
https://github.com/ente-io/photos-app/assets/77285023/e82a3872-cf94-4aad-8a19-7469595440f4

ua741 (approved):

ua741 (dismissed):

ua741 (approved):

PR #1618, opened 2023-12-21
Fix all await warnings

vishnukvmd:


lib/ui/viewer/file/thumbnail_widget.dart
```diff
@@ -220,15 +220,15 @@ class _ThumbnailWidgetState extends State<ThumbnailWidget> {
           _logger.fine("Removing localID reference for " + widget.file.tag);
           widget.file.localID = null;
           if (widget.file.isTrash) {
-            TrashDB.instance.update(widget.file as TrashFile);
+            await TrashDB.instance.update(widget.file as TrashFile);
```
ua741, line 223: This and another line is intentional to avoid blocking UI on this operation. Can mark it as unawaited

ua741 (approved):

ua741 (approved):

PR #1635, opened 2024-01-05
Move from ObjectBox to Isar

vishnukvmd:


lib/db/embeddings_db.dart
```diff
@@ -0,0 +1,65 @@
...
+
+  Future<List<Embedding>> getUnsyncedEmbeddings() async {
+    return await _isar.embeddings.filter().updationTimeEqualTo(null).findAll();
+  }
+
+  Future<void> deleteAllForModel(Model model) async {
+    await _isar.writeTxn(() async {
+      final embeddings =
+          await _isar.embeddings.filter().modelEqualTo(model).findAll();
+      await _isar.embeddings.deleteAll(embeddings.map((e) => e.id).toList());
```
0nullpointer, line 55: Do you need an embedding deleted event similar to `EmbeddingUpdatedEvent`?

lib/db/embeddings_db.dart
```diff
@@ -0,0 +1,65 @@
...
+
+  static final EmbeddingsDB instance = EmbeddingsDB._privateConstructor();
+
+  Future<void> init() async {
+    final dir = await getApplicationDocumentsDirectory();
+    _isar = await Isar.open(
+      [EmbeddingSchema],
+      directory: dir.path,
+    );
+    // TODO: _clearDeprecatedStore(dir);
```
0nullpointer, line 22: Remove?

0nullpointer (approved):

lib/db/embeddings_db.dart
vishnukvmd: Yep.

lib/db/embeddings_db.dart
vishnukvmd: Was hedging, uncommented, will drop the ObjectBox DB.

0nullpointer (approved):

licaon-kter:
So users get migrated from objectbox to isar or how?

Note that F-Droid needs to build isar libs and will reject prebuild .so's :wink:

vishnukvmd:
@licaon-kter we're dropping the data stored in ObjectBox, pulling it from the server, and pushing to Isar.

Will figure out a way to build Isar from source, thanks for the heads up!

PR #1703, opened 2024-02-07
[FEAT] Home Widget for Favorites Album

prateekmedia:
## Description

This PR adds home widget support for both Android for favorites album.

The refresh time of widget slideshow is 10 minutes after which a different random image would be chosen.

Output:

iOS is blocked by workmanager plugin as it doesn't officially supports periodic refresh, but once it does it can be included again by reverting this commit: 96ca7cef81a6057abeb2262b140d2740c92b3f48

Solves #1471


lib/main.dart
```diff
@@ -61,14 +67,128 @@ const kBGPushTimeout = Duration(seconds: 28);
...
+}
+
 void main() async {
   debugRepaintRainbowEnabled = false;
   WidgetsFlutterBinding.ensureInitialized();
   MediaKit.ensureInitialized();
+
+  if (Platform.isAndroid) {
+    try {
+      await Workmanager()
```
ashilkn, line 172: Let's not await on this in `main()` and instead use `.then()` and `.onError()` on the futures? Don't want to increase the time app stays on splash screen.

lib/main.dart
```diff
@@ -61,14 +67,128 @@ const kBGPushTimeout = Duration(seconds: 28);
...
+          .initialize(initSlideshowWidget, isInDebugMode: kDebugMode);
+
+      await Workmanager().registerPeriodicTask(
```
ashilkn, line 175: ^^

lib/ui/tabs/home_widget.dart
```diff
@@ -60,6 +60,8 @@ import 'package:receive_sharing_intent/receive_sharing_intent.dart';
 import "package:shared_preferences/shared_preferences.dart";
 import 'package:uni_links/uni_links.dart';
 
+final scaffoldKey = GlobalKey<ScaffoldState>();
```
ashilkn, line 63: Don't see this used anywhere, lets remove?

lib/ui/tabs/home_widget.dart
```diff
@@ -120,11 +122,13 @@ class _HomeWidgetState extends State<HomeWidget> {
         debugPrint(
           "TabChange going from $_selectedTabIndex to ${event.selectedIndex} souce: ${event.source}",
         );
-        _pageController.animateToPage(
-          event.selectedIndex,
-          duration: const Duration(milliseconds: 100),
-          curve: Curves.easeIn,
-        );
+        if (_pageController.hasClients) {
```
ashilkn, line 125: Why do we need this check?

prateekmedia: It was throwing errors if you turn on Uncaught Exceptions and run photos app.

lib/main.dart
prateekmedia: Done! Although there is analysis check that requires await so I have used unawaited to fix this.

lib/ui/tabs/home_widget.dart
ashilkn: After checking, this exception seems to be coming from `FileSelectionOverlayBar` and not `HomeWidget` so removing this check.

ashilkn (dismissed):

lib/ui/tabs/home_widget.dart
ashilkn: Update: @prateekmedia found that this is required for a different exception, so reverting removal of check.

ashilkn (approved):

PR #1707, opened 2024-02-08
Better update dialog

### LINKED ISSUE #1706: Disable obtrusive update notice/check - Android
Ente for Android displays a notice that is close to full screen when there is an update available with no dismiss option, and no option to disable checking. The notice can be dismissed by tapping somewhere off the notice, although this is not clear. It seems the available options are: update now, or update now:

Screenshot_20240208-071134_ente Photos_1

As new versions can be unstable, if the user prefers to postpone the update, they will be forced to interact with this notice every time they open the app. 

Updating should be optional, the notice should be able to be easily dismissed, and checking for updates should also be optional. For example, I use Obtanium to watch Github for new releases, and I don't require update checks within the app at all.

Ente version 0.8.55
Device Pixel 3
OS LineageOs 20

ashilkn:
## Description

Fixes #1706

### DIFF
--- a/lib/ui/tabs/home_widget.dart
+++ b/lib/ui/tabs/home_widget.dart
@@ -11,6 +11,7 @@ import 'package:media_extension/media_extension_action_types.dart';
 import 'package:modal_bottom_sheet/modal_bottom_sheet.dart';
 import 'package:move_to_background/move_to_background.dart';
 import 'package:photos/core/configuration.dart';
+import "package:photos/core/constants.dart";
 import 'package:photos/core/event_bus.dart';
 import 'package:photos/ente_theme_data.dart';
 import 'package:photos/events/account_configured_event.dart';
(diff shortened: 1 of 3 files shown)


lib/ui/tabs/home_widget.dart
```diff
@@ -197,14 +200,30 @@ class _HomeWidgetState extends State<HomeWidget> {
...
         Future.delayed(Duration.zero, () {
-          showDialog(
-            context: context,
-            builder: (BuildContext context) {
-              return AppUpdateDialog(
-                UpdateService.instance.getLatestVersionInfo(),
-              );
+          SharedPreferences.getInstance().then(
+            (value) {
+              final lastDialogShownTime =
```
vishnukvmd, line 205: Maybe refactor and reuse the code from here?

https://github.com/ente-io/photos-app/blob/d27a70bd37620ab925264f13b1487ff7b2891d46/lib/services/update_service.dart#L76

vishnukvmd (approved):

PR #1728, opened 2024-02-18
Run ML only when device is healthy

vishnukvmd:
## Description
- Setup `MachineLearningController` to fire events that specify whether machine learning `shouldRun` or not
- `shouldRun` is set to `true` only if the device is healthy and the user is not interacting with the app
- Device is deemed to be healthy if it has >= 20% battery and the temperature is <= 36 degrees.

Any future service that has to run ML can hook into these `MachineLearningControlEvent`s to start and pause indexing.

> Note: This PR also moves `semantic_search` into the `machine_learning` package, please ignore that noise.

## Test Plan
- [x] Tested on Android emulator with `SemanticSearchService`


laurenspriem (approved):
Have not specifically tested the branch on my device, but looking through the code this looks good to me.

